PR #12, opened 2013-02-03
Adding conventional PID computation, deprecating current PID computation

jbohren:
This PR adds a member function with two signatures called `Pid::setError()` and adds deprecation warnings to the current `Pid::updatePid()` member functions, along with appropriate documentation.

This new member function is identical to the deprecated functions, except that it implements the conventional  sign of the input to a PID controller (error = target - current) discussed in this issue: https://github.com/willowgarage/ros_controllers/issues/2

### DIFF
--- a/control_toolbox/include/control_toolbox/pid.h
+++ b/control_toolbox/include/control_toolbox/pid.h
@@ -221,6 +221,10 @@ public:
    * function is equivalent to calling \ref setError with negated error
    * arguments.
    *
+   * \deprecated This function assumes <tt> p_error = (state - target) </tt> which is an
+   * unconventional definition of the error. Please use \ref setError instead,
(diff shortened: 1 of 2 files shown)


control_toolbox/include/control_toolbox/pid.h
```diff
@@ -125,14 +125,6 @@ class Pid
...
 
   /*!
-   * \brief Update the Pid loop with nonuniform time step size.
-   *
-   * \param p_error  Error since last call (p_state-p_target)
-   * \param dt Change in time since last call
-   */
-  double updatePid(double p_error, ros::Duration dt);
-
-  /*!
```
jbohren, line 135: This wasn't deleted, it was just moved down near the other updatePid() declaration.

PR #15, opened 2013-02-05
Make public getControllerByName method thread-safe.

adolfo-rt:
Existing virtual non-threadsafe method has been suffixed with -Impl and pushed
to protected class scope. In-class uses call getControllerByNameImpl, as the
lock has already been acquired.


jbohren:
@adolfo-rt thoughts on this fix, instead?

https://github.com/jbohren-forks/ros_control/compare/fix-controllers-rlock

adolfo-rt:
Sure!, a recursive lock is a cleaner solution indeed.

PR #17, opened 2013-02-05
Fixing backwards-compatibility in updatePid() 

jbohren:
This makes the internal computations (like the error term of updatePid() keep the same sign that they did before the API change by bringing back the old updatePid() function contents.

Of course mixing calls to `updatePid()` and `setError()` will mess with the value of the integral gain and the derivative computation (anything that depends on the state), so I also added some warnings about that in the documentation.

This regression was detected by the new tests.

### DIFF
--- a/control_toolbox/include/control_toolbox/pid.h
+++ b/control_toolbox/include/control_toolbox/pid.h
@@ -218,9 +218,7 @@ public:
   double setError(double error, double error_dot, ros::Duration dt);
 
   /*!
-   * \brief Update the Pid loop with nonuniform time step size. NOTE: this
-   * function is equivalent to calling \ref setError with a negated return
-   * value.
+   * \brief Update the Pid loop with nonuniform time step size.  
    *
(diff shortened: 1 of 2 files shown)


wmeeusse:
Can updatePid directly call through to setError, and in that way not create problems when mixing the two methods?

jbohren:
Well, that's what I thought before, but if your error is defined as "state-target" then so will your integrator accumulation. If someone is relying on testing those values (which are accessible), then the current API in master would accumulate in the same direction if you called `setError(e)` as it would if you called `updatePid(-e)`.

I figured the most important thing was preserving the behavior of `updatePid()` so that no code breaks, and because the signs are reversed _inside_ of `updatePid()` when the terms are negated and added up, you can't just get the same effect by negating the arguments. You COULD negate the return value, but then because `updatePid` also sets `_cmd` you would also have to re-set that after calling the function to negate the command that was stored.

wmeeusse:
@jbohren thanks for your comments

PR #23, opened 2013-02-15
Adding lots of inline documentation, rosdoc files

jbohren:
I decided to really sit down and read through all the code, and I figured I'd document it along the way. Please tell me if I misunderstood anything (:

### DIFF
--- a/controller_interface/include/controller_interface/controller.h
+++ b/controller_interface/include/controller_interface/controller.h
@@ -42,6 +42,11 @@
 namespace controller_interface
 {
 
+/** \brief %Controller with a specific hardware interface
+ *
+ * This class is templated by a given type of HardwareInterface
+ *
+ */
 
 template <class T>
 class Controller: public ControllerBase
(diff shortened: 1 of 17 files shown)


controller_manager/include/controller_manager/controller_manager.h
```diff
@@ -57,25 +57,108 @@
...
+   * \param robot_hw A pointer to a robot hardware interface
+   * \param nh The ros::NodeHandle in whose namespace all ROS interfaces should
+   * operate.
+   */
   ControllerManager(hardware_interface::RobotHW *robot_hw,
                    const ros::NodeHandle& nh=ros::NodeHandle());
   virtual ~ControllerManager();
 
-  // Real-time functions
+  /** @name Real-Time Safe Functions
```
adolfo-rt: Could you use \name instead of @name, to be consistent with the used syntax convention?. The same applies to other occurrences of @commands

controller_interface/include/controller_interface/controller.h
```diff
@@ -42,6 +42,11 @@
 namespace controller_interface
 {
 
+/** \brief %Controller with a specific hardware interface
+ *
+ * This class is templated by a given type of HardwareInterface
```
adolfo-rt: How about documenting template parameters with the \tparam command? 
http://www.stack.nl/~dimitri/doxygen/manual/commands.html#cmdtparam

adolfo-rt:
Great job!. Documentation was indeed lacking. The code is largely self-explanatory, but having decent documentation is an indicator that things are getting stable :). Test, code, document.

jbohren:
@adolfo-rt Good call on \tparam

jbohren:
Yeah, I think especially for a system which requires a lot of sub-classing and which is supposed to provide an example for different hardware interfaces, it's good to have API-level doc. I tried to make it as generic to the fundamental design, and avoid kludging it up with implementation details which may change.

PR #24, opened 2013-02-15
Alternative fix to getControllersByName mutex-locking requirements

jbohren:
`ControllerManager::getControllerByName()` requires that the `controllers_lock_` is locked. 

There was a previous fix to this implemented [here](https://github.com/willowgarage/ros_control/pull/15#issuecomment-13596141) but instead of having a public and protected function, it could alternatively be solved by using a reentrant / recursive mutex instead of a unique mutex. This could keep the API a little cleaner, and also allow for more convenience functions like `getControllerByName()`.

### DIFF
--- a/controller_manager/include/controller_manager/controller_manager.h
+++ b/controller_manager/include/controller_manager/controller_manager.h
@@ -161,7 +161,6 @@ public:
   /*\}*/
 
 protected:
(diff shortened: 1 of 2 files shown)


wmeeusse:
Hi @jbohren, thanks again for your contributions. Could you merge this request with the latest documentation changes?

PR #134, opened 2013-12-05
Interface manager

adolfo-rt:
Refs #133.

### DIFF
--- a/hardware_interface/CMakeLists.txt
+++ b/hardware_interface/CMakeLists.txt
@@ -17,6 +17,7 @@ if(USE_ROSBUILD)
   rosbuild_add_gtest(force_torque_sensor_interface_test test/force_torque_sensor_interface_test.cpp)
   rosbuild_add_gtest(imu_sensor_interface_test          test/imu_sensor_interface_test.cpp)
   rosbuild_add_gtest(robot_hw_test                      test/robot_hw_test.cpp)
+  rosbuild_add_gtest(interface_manager_test             test/interface_manager_test.cpp)
 
   # TODO: why is it explicitly needed???, without it the linker fails.
   target_link_libraries(hardware_resource_manager_test     pthread)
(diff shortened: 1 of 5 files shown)


adolfo-rt:
@wmeeusse, please review. Especially 9753aa7. When RobotHW`leverages the`InterfaceManager`class, it internally stores`void_`to interfaces instead of`HardwareInterface_`. Nothing has changed in terms of outside behavior, it's just an implementation detail. Still, I'd like an OK or KO on the proposed changeset.

Being able to store any type, instead of members of a base class is something that I end up using with transmission interfaces.

wmeeusse:
@adolfo-rt Sorry it took me a while to find time to actually look through the source code.  With the storage as void\* and a static cast we loose type checking, but this seems fine as long as we have a guarantee that each type maps on a unique string using the demangle function. Does the demangling also take into account the namespace?  Can you confirm that this mapping is guaranteed to be unique; otherwise this change looks good to go!

adolfo-rt:
> @adolfo-rt Sorry it took me a while to find time to actually look through the source code. With the storage as void\* and a static cast we loose type checking, 

Indeed. That's why I wanted an ack from your part.

> but this seems fine as long as we have a guarantee that each type maps on a unique string using the demangle function. Does the demangling also take into account the namespace? Can you confirm that this mapping is guaranteed to be unique; otherwise this change looks good to go!

Yes, the name demangling is implemented using [gcc intrinsics](https://github.com/ros-controls/ros_control/blob/hydro-devel/hardware_interface/include/hardware_interface/internal/demangle_symbol.h#L52), and does not lose type information (namespaces, template arguments). FYI, when non-gcc compilers are used, we fall back to non-demangled names.

To have an idea of how stored type names look like, you can run the test suite of the `hardware_interface` package and inspect the logs.

```
catkin_make run_tests_hardware_interface
```

These are some examples taken from such output:

```
hardware_interface::ActuatorStateInterface
hardware_interface::JointCommandInterface
hardware_interface::HardwareResourceManager<HandleType, hardware_interface::DontClaimResources>
```

The first two strings are what you should expect to see in the logs as a ros_control user. The last string is an internal type that should never show up in user-facing logs, but serves as an example of a complex type. I took care not to scare users away with unreadable type names ;-)

Thanks for reviewing.

PR #137, opened 2014-01-08
RobotHWGroup Implementation

kphawkins:
I have implemented a RobotHWGroup which seems to match the spec we have been discussing.  Currently, it doesn't support multiple identical interfaces across different RobotHW.  It returns first the locally registered interface, or then one of the interfaces in one of the RobotHW, ordered by when each was registered.


kphawkins:
Unfortunately,  that is not possible. Template methods cannot be made virtual so a call to`get<T>()` inside of ControllerBase would always use the base version. That's why I used this convoluted technique.

wmeeusse:
@kphawkins That makes sense, thanks for the explanation!

kphawkins:
Fixed interface aliasing issue: https://github.com/ros-controls/ros_control/pull/138

bmagyar:
This I believe is now safe to close as #231 addressed this feature. Please open a new one if you have suggestions to improve that implementation.

PR #143, opened 2014-02-19
Add a parameter to configure controller spawner timeout

po1:
This will be very useful, especially for robots which have to go through a long boot sequence before the `controller_manager` is up. On our robot, `30` is not always enough anyway...


adolfo-rt:
+1

PR #148, opened 2014-03-07
Added pos+vel and pos+vel+acc command interfaces

Igorec:
This hardware interface supports commanding joints by position, velocity and acceleration together in one command.

### DIFF
--- a/hardware_interface/CMakeLists.txt
+++ b/hardware_interface/CMakeLists.txt
@@ -68,6 +68,13 @@ else()
 
     catkin_add_gtest(interface_manager_test          test/interface_manager_test.cpp)
     target_link_libraries(interface_manager_test ${catkin_LIBRARIES})
+    
+    catkin_add_gtest(posvel_command_interface_test test/posvel_command_interface_test.cpp)
+    target_link_libraries(posvel_command_interface_test ${catkin_LIBRARIES})
+    
+    catkin_add_gtest(posvelacc_command_interface_test test/posvelacc_command_interface_test.cpp)
+    target_link_libraries(posvelacc_command_interface_test ${catkin_LIBRARIES})
+    
   endif()
 
   # Install
--- a/hardware_interface/include/hardware_interface/internal/hardware_resource_manager.h
+++ b/hardware_interface/include/hardware_interface/internal/hardware_resource_manager.h
@@ -79,6 +79,8 @@ template <class ResourceHandle, class ClaimPolicy = DontClaimResources>
 class HardwareResourceManager : public HardwareInterface, public ResourceManager<ResourceHandle>
 {
 public:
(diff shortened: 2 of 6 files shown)


adolfo-rt:
Great to see that you went all the way and even prepared a PR for the new hardware interface, (with tests!). I'd like to discuss two important issues before a merge:

**Firstly**, in the original discussion at ros-controls/ros_controllers#81, you mention that your usecase requires a hardware interface taking both **position** and **velocity** commands, whereas this PR proposes an interface that not only takes **position** and **velocity**, but also **acceleration** commands. Does your usecase actually need the acceleration command?. I'm not saying that a pos,vel,acc interface would not be useful, but it would seem to me that a pos,vel (no acc) interface would be of much wider use. The downside of using the pos,vel,acc interface for pos,vel (no acc) control is that you need to provide dummy raw acceleration data on handle construction, which can be annoying.

**Secondly**, I think we should brainstorm about a more meaningful name for the interface. The interface is useful in itself even if you're not using splines. How about something along the lines of `PosVelJointInterface`, or `PosVelAccInterface`?.

Igorec:
Thank you for advice!
1. I really require **acceleration** too. But have added another interfaface for only **position** and **velocity**.
2. You are right.  Interface `SplineJointInterface` renamed to  `PosVelAccJointInterface` and added `PosVelJointInterface`.

adolfo-rt:
Great!.

Now that there are two interfaces. Would it be clearer to have two separate header files (and tests)?: `posvel_command_interface.h` `posvelacc_command_interface.h`. This makes it clearer which header file you should include. It's true that the existing `joint_command_interface.h` contains multiple interfaces, but for the sake of clarity, I think that separate headers would be beneficial in this case.

Finally, a few requests to keep the git history clean:
- If you rebase your branch on the current `hydro-devel`, changeset c35b239 will be no longer necessary.
- Squash all commits in a single one, so only the relevant changes that will make it upstream remain.

kphawkins:
Can we merge this commit?  Is there anything else needed?

adolfo-rt:
Thanks for bringing back the attention to this PR. It looks pretty good to me.

One style issue I want to raise is: should the single-command accessors be called `setCommandPosition()` and similar (as it is now) or `setPositionCommand()`. The latter reads out better.

Other than that, the PR would have to be rebased against `indigo-devel` for merging. All new features are being added to this branch. `hydro-devel` is in bugfix-only mode.

It would be great if we could get this merged before ROSCON next week, so these new interfaces can be mentioned in the ros_control talk. I'm pretty busy right now wrapping up the presentation in my free cycles, so if one of you could pick this one up, I'd be grateful.

adolfo-rt:
Closing, as #181 ported this feature to indigo.

adolfo-rt:
...and thanks for your hard work!

PR #158, opened 2014-05-08
controller_manager_tests: Fix library linking

po1:
This is silently ignored by gcc, but clang will refuse to build.


adolfo-rt:
Thanks. I also applied parts of patch that still apply in `indigo-devel`.

PR #165, opened 2014-06-17
[controller_manager_msgs] Add Python helpers to make life easier.

adolfo-rt:
- Helper for getting all active controller managers.
- Helper for determining if a namespace contains the controller manager ROS API.
- Helpers for filtering the output of the 'list_controllers' service by
  type, name, state, hardware_interface and claimed resources.


adolfo-rt:
Not ready for merge yet, needs some tests.

Still, please review if you find these additions useful. I'd especially appreciate remarks on how Pythonic the code is, as I'm not a Python black belt.

This PR was inspired by the existing `rqt_controller_manager`, and `rqt_joint_trajectory_controller` (in the works) plugins, which need to reason about controller managers and controllers. I figured that there was some code that could be shared and factored out.

@kphawkins, @jbohren, @po1

kphawkins:
I think this might be helpful and I can't see any reason not to release if you're happy with it.  It looks Python-esque enough for me. It helps motivate the idea of having separate namespaces for controller managers, which might be helpful for systems with a large number of actuators.

As an aside, I've put together a Python utility file for myself, found here: https://github.com/gt-ros-pkg/universal_robot/blob/hydro-devel-c-api/ur_py_utils/src/ur_py_utils/ur_controller_manager.py
It's a rewrite of the `controller_manager_interface.py` as a class which holds onto the services for more responsive activation.  If someone were interested in copying the code for the class ControllerManager into an official release, that'd be cool.

adolfo-rt:
Interesting. I decided to run some numbers to quantify the gains of calling services one way or another. I chose the `list_controllers` service, whose actual computation is lightweight. The code is in this [Gist](https://gist.github.com/adolfo-rt/a5881e35ac146177fbe2). Results for a single-machine scenario (loopback) and a distributed scenario (wifi, ping time ~2ms) are as follows:

|  | loopback | wifi |
| --- | --- | --- |
| wait for service + create proxy + call service | 4 ms | 34 ms |
| only call service | 2.5 ms | 20 ms |
| only call _persistent_ service | 0.3 ms | 3 ms |

(times are approximate)

The difference between the first and last cases is an order of magnitude in both scenarios. I try to favor persistent services whenever they might be called frequently, and use a simple [reconnection policy](https://github.com/pal-robotics/ros_controllers/blob/rqt-jtc/rqt_joint_trajectory_controller/src/rqt_joint_trajectory_controller/joint_trajectory_controller.py#L499-L507).

These changes would have to be incorporated in a ros-distro transition, as they break the existing API of the Python interface. Now could be a good time, but unfortunately I can't do it. My list of pending indigo tasks is not empty.

adolfo-rt:
There go the tests. I'll squash a few commits before merging, but I'll keep them separate for now to ease code review.

adolfo-rt:
So I was adding a `<test_depends>unittest</test_depends>` based on apparently outdated doc like this [wiki page](http://wiki.ros.org/unittest#Update_your_manifest) and this [question](http://answers.ros.org/question/115526/is-test_depend-actually-used-for-anything/), and the Travis build failed with:

```
ERROR: the following packages/stacks could not have their rosdep keys resolved to system dependencies:
controller_manager_tests: Cannot locate rosdep definition for [unittest]
```

I have opted for removing the `unittest` dep for now because the `rostest` dep should transitively take care of it, but I wonder what is the right way to make it explicit. Before moving the question further upstream, does any of the issue readers know if and how the `unittest` dep should be added?.

po1:
> I have opted for removing the unittest dep for now because the rostest dep should transitively take care of it, but I wonder what is the right way to make it explicit. Before moving the question further upstream, does any of the issue readers know if and how the unittest dep should be added?.

I believe `unittest` is part of the python standard library. No package is needed, and as such, it doesn't have a rosdep rule.

po1:
Also, nice work here. I'm a bit late to the party, but it looks pretty good.
About the speed of using services, I would like to see how much time firing a new python interpreter actually takes. I made a very quick test some time ago here: pal-robotics/play_motion#37.
My test used the `rosservice` CLI tool, but it might do additional things that your implementation avoids. The results indicated that calling `rosservice call` from the shell takes nearly a second, which seems a lot compared to your results.

adolfo-rt:
Squashed commits for merge.

PR #170, opened 2014-06-24
Restore hydro for CI.

adolfo-rt:

PR #175, opened 2014-07-27
make rostest in CMakeLists optional (ros/rosdistro#3010)

bulwahn:


bulwahn:
As the same change passed in the hydro-devel branch (https://github.com/ros-controls/ros_control/pull/176), and I tested it with trusty on my local machine, I assume that the CI would pass once it is fixed.
If anyone has doubts, please re-test this commit.

adolfo-rt:
CI builds are broken for Indigo because the Travis build environment uses 12.04, so we can't install ROS binaries. See #170 for more details. For now I've decided to simply wait for Travis CI to support 14.04.

Anyway, the patch looks fine, and AFAIU the rostest dependency is used only for pulling cmake macros.

Thanks!.

PR #191, opened 2014-11-03
Update package maintainers

adolfo-rt:
@wmeeusse, @vpradeep07, I'm assuming that you don't want to remain as maintainers of these packages. Let me know if you think otherwise.


vpradeep07:
Good, call.  Thanks!

LGTM

Issue #193, opened 2014-12-05
PositionJointSaturationHandle and PositionJointSoftLimitsHandle are stateful


mathias-luedtke:
Both position limit handles store the previous command internally.
This leads to problems with mode switching, since the position command might not be updated while velocity or effort joint interfaces are used for control.

I currently see three solutions:
1. implement a reset method, that sets prev_cmd_ to NaN (as in the constructor), this requires some external logic to work properley (e.g. decide when to reset)
2. Get rid of state and limit command based on actual position rather than last command. (Are there any diadvantages with this approach?)
3. Hybrid solution: Introduce a flag (or template code) to have both alternatives.

Do you have any comments?

Currently we have a work-around with a large period that updates prev_cmd_, but we would prefer a proper solution.

adolfo-rt:
Your observations are correct.

The current open-loop implementation needs a `reset()` method because it is stateful. Apart from being useful when mode switching, it is also relevant when recovering from an e-stop. I actually implemented this behavior on an in-house specialization of the class because one of our platforms needed it.

The first implementation I did was indeed closed-loop (used actual pos/vel), but on some platforms the control delay could be too slow to make this reactive enough. The `reset()` method is meaningless in the closed-loop implementation, as it's stateless.

In the end, I think having both implementations available would be valuable. If your control delay is low enough, the closed loop implementation is definitely the better choice.

I haven't thought much about it, but if you'd like to discuss supporting both (with minimal API changes if possible), count me in.

This is the changeset where the implementations in question became open loop: https://github.com/ros-controls/ros_control/commit/2a66843bd0b2c16761386fb3f3de4895cc49d111
In retrospective, I'm not fond of not having simply supported both alternatives from the beginning.

mathias-luedtke:
The closed-loop implementation didn't work for us either, in our case it somehow resulted in over-limiting.
However, it could be achieved by calling reset before calling enforceLimits.

adolfo-rt:
I'm pretty sure that you are observing over-limiting because your control delay is too large for the closed-loop implementation to work. This was precisely the reason why the open-loop variant was implemented.

I presume that you see over-limitation in the maximum velocity. The way velocity limits are enforced in closed-loop goes like this:
- Approximate the velocity for reaching the next command as `(cmd_pos - curr_pos) / period`.
- If this value exceeds the maximum joint velocity, then saturate the command.

[...]

PR #194, opened 2014-12-12
reset functionality for stateful position limit handles

mathias-luedtke:
As discussed in #193 I have added reset functions to the stateful position limit handles.
Existing API was not touch, I have just added new methods to the handles and the corresponding interfaces.

### DIFF
--- a/joint_limits_interface/include/joint_limits_interface/joint_limits_interface.h
+++ b/joint_limits_interface/include/joint_limits_interface/joint_limits_interface.h
@@ -115,6 +115,13 @@ public:
     prev_cmd_ = cmd;
   }
 
+  /**
+   * \brief Reset state, in case of mode switch or e-stop
+   */
+  void reset(){
+    prev_cmd_ = std::numeric_limits<double>::quiet_NaN();
+  }
+
 private:
   hardware_interface::JointHandle jh_;
   JointLimits limits_;
@@ -237,6 +244,13 @@ public:
 
     // Cache variables
     prev_cmd_ = jh_.getCommand();
+  }
+
+  /**
+   * \brief Reset state, in case of mode switch or e-stop
+   */
+  void reset(){
+    prev_cmd_ = std::numeric_limits<double>::quiet_NaN();
   }
 
 private:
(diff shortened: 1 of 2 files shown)


joint_limits_interface/include/joint_limits_interface/joint_limits_interface.h
```diff
@@ -554,10 +568,38 @@ class JointLimitsInterface : public hardware_interface::ResourceManager<HandleTy
...
 /** Interface for enforcing limits on a position-controlled joint with soft position limits. */
-class PositionJointSoftLimitsInterface : public JointLimitsInterface<PositionJointSoftLimitsHandle> {};
+class PositionJointSoftLimitsInterface : public JointLimitsInterface<PositionJointSoftLimitsHandle> {
+public:
+  /** \name Real-Time Safe Functions
+   *\{*/
+  /** \brief Reset all managed handles. */
+  void reset()
+  {
+    typedef hardware_interface::ResourceManager<PositionJointSoftLimitsHandle>::ResourceMap::iterator ItratorType;
```
adolfo-rt, line 595: I think strict compilers will complain of a missing `typename` here. It should read `typedef typename hardware_interface...`

The above statement designates a nested type of a templated class, and compilers need help in figuring out that we're actually talking about a type here.

adolfo-rt:
PR looks good except for one minor comment.

Could you also please extend the test suite to exercise the `reset()` methods, both for handles and interfaces?.

mathias-luedtke:
I have added tests for bothinterfaces. The corresponding handles are tested implicitly.
Actually clang++complains if I add "typename", it is no dependent type name anymore.

adolfo-rt:
I hope to give your tests a spin later today and propose a modification. As they stand, the new tests are basically checking that `reset()` compiles, but are not really checking that it does what it's supposed to. The changes to the test should be minor in implementation effort, but important in terms of validating what they should.

adolfo-rt:
I take my last comment back. The tests are fine.

Thanks so much for the contribution with tests!

PR #205, opened 2015-08-13
Address -Wunused-parameter warnings

adolfo-rt:
Building with stricter gcc flags produces many warnings. This PR addresses an important part of those.


adolfo-rt:
No news, good news. Merging.

PR #200, opened 2015-03-18
HW interface switch feature with unit tests

mathias-luedtke:
As promised in  #197 I have implemented a new version of the HW interface switch.

The RobotHW class got new members as discussed beforehand:

``` c++
virtual bool canSwitch(const std::list<ControllerInfo> &start_list,
                       const std::list<ControllerInfo> &stop_list) const;

virtual void doSwitch(const std::list<ControllerInfo> &start_list,
                      const std::list<ControllerInfo> &stop_list);
```

These should check and execute the switches on the hardware.

The `switchController` filters the start and stop list as follows:
- filter all restarts (make start_list and stop_list disjoint)
- filter stops of stopped controllers (raise an error for strict checking)
- filter starts (not restarts) of started controllers (raise an error for strict checking)

I have put a lot of effort into the unit tests, they should cover most cases, including the filters, resource and inter-controller mode conflicts.
The SwitchBot only allows direct (in one function call) switches if new_hwi >= old_hwi (string comparison).

@ipa-fxm will implement a gazebo plugin again, I will test it with https://github.com/ipa320/ros_canopen.

PR #206, opened 2015-08-13
transmission_interface: Add new loading method

adolfo-rt:
Allow loading transmissions from a vector of TransmissionInfo instances.

**Motivation**
This is irrelevant to the PR itself, but describes why we considered the proposed addition useful.

We have added support in our robots to change joint control modes at runtime (leveraging #200), and for that to work, we needed to reason about which actuators are associated to each joint requesting a control mode switch. Note that sometimes multiple actuators are coupled to a single joint, or the other way around.

All the information required to perform this reasoning is contained in the `vector<TransmissionInfo>`  that results from parsing a URDF. Before this PR, this vector was a implementation detail of the `TransmissionInterfaceLoader` class, and was not accessible to the outside. The proposed overload allows a user to generate the `vector<TransmissionInfo>`, and then:
- On initialization: Pass it to the `TransmissionInterfaceLoader`, for populating transmission and joint interfaces.
- When the robot is running: Keep it for reasoning about joint control mode switching.

### DIFF
--- a/transmission_interface/include/transmission_interface/transmission_interface_loader.h
+++ b/transmission_interface/include/transmission_interface/transmission_interface_loader.h
@@ -374,6 +374,16 @@ public:
   bool load(const std::string& urdf);
 
   /**
+   * \brief Load all transmissions contained in a vector of \c TransmissionInfo instances.
+   *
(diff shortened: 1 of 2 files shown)


adolfo-rt:
Going ahead here...

PR #207, opened 2015-08-17
Per controller update period

adolfo-rt:
Superceeds #127.

I'm proposing the feature in jade first, as the tests leverage the testing infrastructure I put in place for #204. ~~Because of that, please ignore all commits until 5c3f066, which is when this feature actually starts~~.

The changes are basically what @meyerj initially proposed, with some variable rename proposals, minor code refactoring and tests. I'd like some feedback especially on the renaming done in c0c38fd and 3ddb2f0. Once we have an agreement on that, I'll squash the changesets and backport to indigo.

### DIFF
--- a/controller_interface/include/controller_interface/controller_base.h
+++ b/controller_interface/include/controller_interface/controller_base.h
@@ -108,7 +108,7 @@ public:
       starting(time);
       state_ = RUNNING;
       skipped_update_cycles_ = 0;
-      total_update_period_ = ros::Duration();
+      time_since_last_update_ = ros::Duration();
       return true;
     }
     else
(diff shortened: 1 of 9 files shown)


controller_interface/include/controller_interface/controller_base.h
```diff
@@ -36,6 +36,12 @@
 #include <ros/node_handle.h>
 #include <hardware_interface/robot_hw.h>
 
+namespace hardware_interface
+{
+  class InterfaceResources;
+}
+
+namespace controller_manager { class ControllerManager; }
```
fmessmer: style inconsistency for forward-declarations

adolfo-rt: Fixed. Good catch

jacknlliu:
Hi, what's the status of this PR, any update?

MaximilienNaveau:
Hi, I am very interested in this PR as well, I would love to know if this will be worked on.
Thanks

dsandber:
ROS2 seems to support this: https://github.com/ros-controls/ros2_control/issues/872

bmagyar:
Thank you @dsandber , indeed it is supported in ROS2. Guys I'm really sorry but none of us have any bandwidth to support ROS Noetic anymore. Please come over to the ROS2 world bit make sure to avoid Python launch files ;)

PR #209, opened 2015-09-04
controller_manager: Fix doSwitch execution point

adolfo-rt:
The doSwitch method needs to be executed in the update() method,  that is, in
the real-time path, which is where controller switching actually takes place.

It was previously done in the switchController callback, which is non real-time.
In this method controller switching is scheduled, but not actually executed.

This changeset fixes a bug in which hardware interface  modes could switch
before controllers, leading to undefined behavior.

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -94,6 +94,9 @@ void ControllerManager::update(const ros::Time& time, const ros::Duration& perio
   // there are controllers to start/stop
   if (please_switch_)
   {
+    // switch hardware interfaces (if any)
+    robot_hw_->doSwitch(switch_start_list_, switch_stop_list_);
+
     // stop controllers
     for (unsigned int i=0; i<stop_request_.size(); i++)
       if (!stop_request_[i]->stopRequest(time))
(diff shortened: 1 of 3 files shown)


controller_manager/src/controller_manager.cpp
```diff
@@ -408,24 +411,37 @@ bool ControllerManager::switchController(const std::vector<std::string>& start_c
...
+  switch_start_list_.clear();
+  switch_stop_list_.clear();
   
   std::vector<ControllerSpec> &controllers = controllers_lists_[current_controllers_list_];
   for (size_t i = 0; i < controllers.size(); ++i)
   {
     bool in_stop_list  = false;
-
     for(size_t j = 0; j < stop_request_.size(); j++)
-      in_stop_list = in_stop_list || (stop_request_[j] == controllers[i].c.get());
```
adolfo-rt, line 419: This is a small optimization that allows to exit the loop earlier, without having to iterate through the whole range. Same thing a few lines below.

mathias-luedtke:
Please do not merge this yet, it breaks ros_canopen.

adolfo-rt:
Ack. Keep us posted.

adolfo-rt:
Updated to build on top of #217. `canSwitch`→`prepareSwitch`. Tests are passing.

Issue #211, opened 2015-09-04
[controller_manager] Fix execution point of doSwitch


adolfo-rt:
When switching controllers, the robot hardware abstraction might require changing the control mode of certain resources (e.g. joints). This feature was contributed in #200. I recently realized that the current implementation has an important flaw with a straightforward fix.

When a controller switch request is made, a non real-time ROS callback checks that the switch is possible, and among other things calls [`RobotHW::canSwitch(...)`](https://github.com/ros-controls/ros_control/blob/indigo-devel/controller_manager/src/controller_manager.cpp#L475). This callback doesn't perform the actual switching, but [delegates it](https://github.com/ros-controls/ros_control/blob/indigo-devel/controller_manager/src/controller_manager.cpp#L95-L108) to the [real-time] control thread. Because of this, the call to `RobotHW::doSwitch(...)` should also be done in the control thread. We currently do it in the [ROS callback](https://github.com/ros-controls/ros_control/blob/indigo-devel/controller_manager/src/controller_manager.cpp#L483), which is wrong.

This misplaced call to `RobotHW::doSwitch(...)` can cause a robot to switch the control mode of its joints before the new controllers are actually running. The ensuing behavior is undefined, and depends on how your hardware abstraction is implemented. I could reproduce this on hardware, and confirm that the fix proposed below address the issue.

The fix consists of moving the execution point of `RobotHW::doSwitch(...)`, which entails that the method implementation must now be **real-time safe**. A fix for indigo is proposed in #209.

Because the real-time safe requirements now imposed on `RobotHW::doSwitch(...)`, it now becomes much likelier that `RobotHW::canSwitch(...)` will have side-effects, i.e., change the state of the `RobotHW` instance. For this reason, I additionally propose that in jade we [remove the constness](https://github.com/ros-controls/ros_control/commit/f19abeabfbd58ec986de507099ba62ef9291d778) of the `RobotHW::canSwitch(...)` method. Since jade has not been officially released, we can allow to make this small API change. A fix for jade is porposed in #210.

@ipa-mdl

mathias-luedtke:
You are right it might lead to undefined behaviour if this case is not handled properly in RobotHW, which is rather cumbersome.
But mode switching might take longer than 1 update cycle, so I am not sure how to integrate it in `update()` properly.

For CANopen it takes 2 to 10 (rough estimation) cycles to switch the mode depending on the device, which might require to switch the motor power off and on again.
[...]

PR #214, opened 2015-10-27
Introduce prepareSwitch

mathias-luedtke:
This PR introduces `prepareSwitch` as a replacement for `canSwitch`, which will be deprecated.
The reasons behind are discussed in #211.

Summary:
`doSwitch` should be run in the RT-part (#209,#210), so it must not perform blocking tasks.
Thus much more work needs to be done in `canSwitch`. That's why it is preferable to have a non-const equivalent.  A new name was applied to emphasize the extended role.

Existing implementations with `canSwitch` will  still work.

### DIFF
--- a/hardware_interface/include/hardware_interface/robot_hw.h
+++ b/hardware_interface/include/hardware_interface/robot_hw.h
@@ -101,8 +101,16 @@ public:
    * Check (in non-realtime) if given controllers could be started and stopped from the current state of the RobotHW
    * with regard to necessary hardware interface switches. Start and stop list are disjoint.
    * This is just a check, the actual switch is done in doSwitch()
+   * @deprecated: Implement prepareSwitch() instead
    */
   virtual bool canSwitch(const std::list<ControllerInfo>& /*start_list*/, const std::list<ControllerInfo>& /*stop_list*/) const { return true; }
+
+  /**
+   * Check (in non-realtime) if given controllers could be started and stopped from the current state of the RobotHW
+   * with regard to necessary hardware interface switches and prepare the switching. Start and stop list are disjoint.
+   * This handles the check and preparation, the actual switch is commited in doSwitch()
+   */
+  virtual bool prepareSwitch(const std::list<ControllerInfo>& start_list, const std::list<ControllerInfo>& stop_list) { return canSwitch(start_list, stop_list); }
 
   /**
    * Perform (in non-realtime) all necessary hardware interface switches in order to start and stop the given controllers.
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -472,7 +472,7 @@ bool ControllerManager::switchController(const std::vector<std::string>& start_c
     return false;
   }
 
-  if (!robot_hw_->canSwitch(start_list, stop_list))
+  if (!robot_hw_->prepareSwitch(start_list, stop_list))
   {
     ROS_ERROR("Could not switch controllers. The hardware interface combination for the requested controllers is unfeasible.");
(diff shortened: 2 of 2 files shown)


hardware_interface/include/hardware_interface/robot_hw.h
```diff
@@ -101,8 +101,16 @@ class RobotHW : public InterfaceManager
    * Check (in non-realtime) if given controllers could be started and stopped from the current state of the RobotHW
    * with regard to necessary hardware interface switches. Start and stop list are disjoint.
    * This is just a check, the actual switch is done in doSwitch()
+   * @deprecated: Implement prepareSwitch() instead
    */
-  virtual bool canSwitch(const std::list<ControllerInfo>& /*start_list*/, const std::list<ControllerInfo>& /*stop_list*/) const { return true; }
+  virtual bool canSwitch(const std::list<ControllerInfo>& /*start_list*/, const std::list<ControllerInfo>& /*stop_list*/) { return true; }
```
adolfo-rt: Nit: Also add `ROS_DEPRECATED` before method ([doc](http://wiki.ros.org/CppStyleGuide#Deprecation)).

adolfo-rt: We can actually start printing warnings about the deprecation in jade, not indigo.

adolfo-rt:
Picking this one up.

@ipa-mdl. Please confirm if the following course of action is OK for your usecase:
1. Merge this PR into `indigo-devel`.
2. Rebase #209 on top of this PR, merge into `indigo-devel` (@ipa-mdl **confirm this is fine for you**).
3. Forward-port this PR to `jade-devel`, adding `ROS_DEPRECATED` warning (`canSwitch` will dissapear in ROS kinetic).
4. Rebase #210 on top of the jade port of this PR, merge into `jade-devel`.
5. Rollout releases of `ros_control` into indigo and jade (finally).

adolfo-rt:
lgtm
(by the way)

hardware_interface/include/hardware_interface/robot_hw.h
mathias-luedtke: Should I remove the `@deprecated` tag in indigo?
If we add `ROS_DEPRECATED` we might get deprecation warnings while building `controller_manager` (I have to check), but users implementing canSwitch will not get any warning.

mathias-luedtke:
The course is fine for us, I have already implemented the new execution point behaviour in https://github.com/ros-industrial/ros_canopen/pull/150

I can rebase and test #209, but we do not support jade for now.

hardware_interface/include/hardware_interface/robot_hw.h
mathias-luedtke: Update:
g++ will print plenty of deprecation warnings, especially for every controller because `controller_base.h` includes `robot_hw.h`.
clang++ is less noisy, it warns only if `canSwitch` is accessed directly.

So I would suggest to not add `ROS_DEPRECATED` in this special case.

adolfo-rt:
This is great, will proceed then.

hardware_interface/include/hardware_interface/robot_hw.h
adolfo-rt: All right, we will mention it prominently in the release statement then.

mathias-luedtke:
I have just noticed that I removed `const` from `canSwitch`. This was not intended.
Fixed in 172e0fd89598abfe3deaad5bb61f420e68d6f081.

adolfo-rt:
Please check #217. It's the changes proposed by @ipa-mdl, with a more descriptive commit message.

PR #220, opened 2015-11-30
Allow the InterfaceManager class to register other InterfaceManagers.

toliver:
This will make it possible to combine several RobotHW objects into a single one.

This PR contains the part of #138 that deals with combining interface_managers. (Also partially related to  issues discussed in #75 and #151).

To deal with the memory leak mentioned [here](https://github.com/ros-controls/ros_control/pull/138#issuecomment-48914503), I've used a `boost::ptr_vector<ResourceManagerBase>` that will deal with the destruction of any combined `ResourceManager` when the InterfaceManager is destroyed.

This approach allows us to keep the signature of `T* get()` (not breaking compatibility) but still address the issues that @kphawkins expressed [here](https://github.com/ros-controls/ros_control/pull/138#issuecomment-52702242). This means that any pointer returned by `T* get()` would stay valid until the InterfaceManager is destroyed, even if a subsequent call to `T* get()` for the same interface T has generated a new combined interface object. In this case, the user of the old pointer would simply have a shorter list of resources available.

### DIFF
--- a/hardware_interface/include/hardware_interface/internal/interface_manager.h
+++ b/hardware_interface/include/hardware_interface/internal/interface_manager.h
@@ -26,20 +26,60 @@
 // POSSIBILITY OF SUCH DAMAGE.
 //////////////////////////////////////////////////////////////////////////////
 
-/// \author Wim Meussen, Adolfo Rodriguez Tsouroukdissian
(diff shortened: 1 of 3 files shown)


toliver:
Reopen against kinetic-devel as #235

PR #224, opened 2015-12-09
Switch TravisCI to run on Trusty and Indigo

otamachan:
Travis supports Trusty now(https://docs.travis-ci.com/user/trusty-ci-environment/).
Switched Travis CI to use on Trusty and Indigo.

This seems to fix CI build. https://travis-ci.org/otamachan/ros_control/builds/95798345

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 # Travis Continuous Integration Configuration File For ROS Control Projects
 # Author: Dave Coleman
+sudo: required
+dist: trusty
 language:
   - cpp
   - python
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -42,7 +44,7 @@ install: # Use this to install any prerequisites or dependencies necessary to ru
   # Install dependencies for source repos
   - rosdep install --from-paths src --ignore-src --rosdistro $ROS_DISTRO -y
 before_script: # Use this to prepare your build for testing e.g. copy database configurations, environment variables, etc.
+  - export PYTHONPATH=/usr/lib/python2.7/dist-packages:$PYTHONPATH
   - source /opt/ros/$ROS_DISTRO/setup.bash  
 script: # All commands must exit with code 0 on success. Anything else is considered failure.
-  - catkin_make -j2
-
+  - catkin_make -j2 --cmake-args -Wno-dev
```
davetcoleman: What does Wno-dev do?

davetcoleman:
This is great, could you add it to ros_controllers too?

.travis.yml
efernandez: From `man cmake`:

```
       -Wno-dev
              Suppress developer warnings.

              Suppress  warnings  that are meant for the author of the CMakeLists.txt
              files.
```

@otamachan Is this really needed? Which warnings do you see?

bmagyar: I'd prefer to have all warnings if they make sense.

otamachan: Without `-Wno-dev`, it generates a lots of warnings.
https://travis-ci.org/otamachan/ros_control/builds/95517583#L2152

It seems that it's because cmake version in TravisCI is 3.2.2.
I think It's better to add cmake_policy in catkin(https://github.com/ros/catkin) rather than in this repo.

davetcoleman: > I think It's better to add cmake_policy in catkin(https://github.com/ros/catkin) rather than in this repo.

I don't know what you mean by that, but perhaps open an issue in the catkin repo explaining what needs to be done?

I think I'd rather leave those warnings there since the point of Travis is to catch problems with the code

otamachan:
I removed `-Wno-dev`.

I'm testing for ros_controllers now.

.travis.yml
mathias-luedtke: This is a bug in catkin (https://github.com/ros/catkin/issues/761). It is already fixed, but not yet released.

PR #231, opened 2016-01-21
CombinedRobotHW implementation

toliver:
(Includes changes in PR #220)

This is related to some of the issues discussed in #75.

CombinedRobotHW allows to dynamically load a list of RobotHW objects defined in the parameter server (using pluginlib à la controller manager) and combines them into a single RobotHW. A single controller manager can then access resources from any robot.

Configuration example: 

``` yaml
robot_hardware:
- my_robot_hw_1
- my_robot_hw_2
- right_arm
- left_arm
my_robot_hw_1:
  type: combined_robot_hw_tests/MyRobotHW1
my_robot_hw_2:
  type: combined_robot_hw_tests/MyRobotHW2
  joints:
  - test_joint4
  - test_joint5
right_arm:
  type: my_arm/MyArmHW
  address: 192.168.1.100
  robot_description: robot_description
  joint_name_filter: ra_
left_arm:
  type: my_arm/MyArmHW
  address: 192.168.1.101
  robot_description: robot_description
  joint_name_filter: la_
```

This is a similar approach to #151 but removing the need for a separate DeviceHW class, and for a URDF to be passed. It uses #220 to combine the resources of the different RobotHWs.

~~Only one change in this PR is not backwards compatible: All the classes derived from RobotHW will have to declare a parameterless constructor in order to build correctly.~~ (See https://github.com/ros-controls/ros_control/pull/231#issuecomment-173933896)

If a RobotHW derived class is intended to be loaded by CombinedRobotHW then it needs to implement at least:
- Parameterless constructor
- `bool init(ros::NodeHandle& root_nh, ros::NodeHandle &robot_hw_nh)` function that sets up and initializes the RobotHW (potentially reading ROS parameters)
- Parameterless `read()` and/or `write()`

Our motivation to write this class is to have an easy way to combine the Shadow Robot hand with different types of arms, so that we can have access to all the robot hardware resources from a single controller (e.g. to implement an impedance controller for arm+hand).

### DIFF
--- a/combined_robot_hw/include/combined_robot_hw/combined_robot_hw.h
+++ b/combined_robot_hw/include/combined_robot_hw/combined_robot_hw.h
@@ -1,0 +1,121 @@
+///////////////////////////////////////////////////////////////////////////////
+// Copyright (C) 2015, Shadow Robot Company Ltd.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+// * Redistributions of source code must retain the above copyright notice,
+// this list of conditions and the following disclaimer.
+// * Redistributions in binary form must reproduce the above copyright
+// notice, this list of conditions and the following disclaimer in the
+// documentation and/or other materials provided with the distribution.
+// * Neither the name of Shadow Robot Company Ltd. nor the names of its
+// contributors may be used to endorse or promote products derived from
+// this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+//////////////////////////////////////////////////////////////////////////////
+
+#ifndef COMBINED_ROBOT_HW_COMBINED_ROBOT_HW_H
+#define COMBINED_ROBOT_HW_COMBINED_ROBOT_HW_H
+
+#include <list>
+#include <map>
+#include <typeinfo>
+#include <hardware_interface/internal/demangle_symbol.h>
+#include <hardware_interface/internal/interface_manager.h>
+#include <hardware_interface/hardware_interface.h>
+#include <hardware_interface/robot_hw.h>
+#include <pluginlib/class_loader.h>
+#include <ros/console.h>
+#include <ros/node_handle.h>
+
+namespace combined_robot_hw
+{
+
+/** \brief CombinedRobotHW
+ *
+ * This class provides a way to combine RobotHW objects.
+ *
+ *
+ *
+ */
+class CombinedRobotHW : public hardware_interface::RobotHW
+{
+public:
+  CombinedRobotHW();
+
+  virtual ~CombinedRobotHW(){}
+
+  /** \brief The init function is called to initialize the RobotHW from a
+   * non-realtime thread.
+   *
+   * \param root_nh A NodeHandle in the root of the caller namespace.
+   *
+   * \param robot_hw_nh A NodeHandle in the namespace from which the RobotHW
+   * should read its configuration.
+   *
+   * \returns True if initialization was successful
+   */
+  virtual bool init(ros::NodeHandle& root_nh, ros::NodeHandle &robot_hw_nh);
+
+
+  /**
+   * Check (in non-realtime) if given controllers could be started and stopped from the current state of the RobotHW
+   * with regard to necessary hardware interface switches and prepare the switching. Start and stop list are disjoint.
+   * This handles the check and preparation, the actual switch is commited in doSwitch()
+   */
+  virtual bool prepareSwitch(const std::list<hardware_interface::ControllerInfo>& start_list,
+                             const std::list<hardware_interface::ControllerInfo>& stop_list);
+
+  /**
+   * Perform (in realtime) all necessary hardware interface switches in order to start and stop the given controllers.
+   * Start and stop list are disjoint. The feasability was checked in prepareSwitch() beforehand.
+   */
+  virtual void doSwitch(const std::list<hardware_interface::ControllerInfo>& start_list,
+                        const std::list<hardware_interface::ControllerInfo>& stop_list);
+
+  /**
+   * Reads data from the robot HW
+   *
+   * \param time The current time
+   * \param period The time passed since the last call to \ref read
+   */
+  virtual void read(const ros::Time& time, const ros::Duration& period);
+
+  /**
+   * Writes data to the robot HW
+   *
+   * \param time The current time
+   * \param period The time passed since the last call to \ref write
+   */
+  virtual void write(const ros::Time& time, const ros::Duration& period);
+
+protected:
+  ros::NodeHandle root_nh_;
+  ros::NodeHandle robot_hw_nh_;
+  pluginlib::ClassLoader<hardware_interface::RobotHW> robot_hw_loader_;
+  std::vector<boost::shared_ptr<hardware_interface::RobotHW> > robot_hw_list_;
+
+  virtual bool loadRobotHW(const std::string& name);
+
+  /** \brief Filters the start and stop lists so that they only contain the controllers and
+   * resources that correspond to the robot_hw interface manager
+   */
+  void filterControllerList(const std::list<hardware_interface::ControllerInfo>& list,
+                            std::list<hardware_interface::ControllerInfo>& filtered_list,
+                            boost::shared_ptr<hardware_interface::RobotHW> robot_hw);
+};
+
+}
+
+#endif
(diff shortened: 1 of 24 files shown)


carlosjoserg:
It sounds good, like I would want to use it a lot... my humble opinion as an user, if I may...
- Naming is always one of the hardest tasks. I think I prefer the class name to be `RobotHWGroup` from #138 or something like `RobotHWCollector` or `RobotHWManager`? Though I'm sure the community can propose more interesting names.
- I never liked the `joint_name_filter:` parameter. Perhaps first time I saw it was [here](https://github.com/ipa320/cob_gazebo_plugins/tree/indigo_dev/cob_gazebo_ros_control) for a gazebo plugin. I typically write the hw for a particular robot which I _link_ to an urdf model. So the joint names are known by design. Thus, I think each hwiface should be aware of its own joints from a _combined_ `robot_description` parameter using the robot name, or namespace, or prefix (used in ROS-I). But probably is my way of thinking...

> Parameterless read() and/or write()
- If one uses the [`joint_limits_interface`](https://github.com/ros-controls/ros_control/blob/jade-devel/joint_limits_interface/include/joint_limits_interface/joint_limits_interface.h), one needs to pass the period, or elapsed time, computed from the control loop, as it is done e.g. in the `gazebo_ros_control` plugin [here](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/jade-devel/gazebo_ros_control/src/default_robot_hw_sim.cpp#L263-L285) or in the (unofficial but useful) `ros_control_boilerplate` template [here](https://github.com/davetcoleman/ros_control_boilerplate/blob/jade-devel/rrbot_control/src/rrbot_hw_interface.cpp#L64-L67)

combined_robot_hw/include/combined_robot_hw/combined_robot_hw.h
```diff
@@ -0,0 +1,115 @@
...
+// Copyright (C) 2015, Shadow Robot Company Ltd.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+// * Redistributions of source code must retain the above copyright notice,
+// this list of conditions and the following disclaimer.
+// * Redistributions in binary form must reproduce the above copyright
+// notice, this list of conditions and the following disclaimer in the
+// documentation and/or other materials provided with the distribution.
+// * Neither the name of PAL Robotics S.L. nor the names of its
```
carlosjoserg: [minor] PAL or Shadow?

bmagyar:
Hi,
I will have to take some time to give this a shot. Would you consider adding a Gazebo example too?

I understand that this would be very desired to use with modular robot products such as Shadow's robots or PAL's TIAGo.

Let me begin with the quick and simple but annoying notes.
- All commits on the PR should compile. Travis gives a weird dependency error with the first one here.
- Please refrain from merging remotes on top of the PR branch. The best practice is to rebase your changes upon the current state of the repository, so that your modifications are added in a clean way rather than being sewn into the history of the repository.

Are you sure that API changes are utmost necessary? I'm not sure how the jade release would tolerate breaking API. Is it still young enough to allow for it, @davetcoleman ?

mathias-luedtke:
> All commits on the PR should compile. Travis gives a weird dependency error with the first one here.

The dependency error results from the broken travis config. 

> The best practice is to rebase

This should solve the dependency error ;)

toliver:
@carlosjoserg 
- I agree that the name is not the nicest. I like `RobotHWGroup`, although `CombinedRobotHW` kind of stresses the fact that it is itself a RobotHW child. Let's see if somebody comes up with an even better option. Otherwise changing to `RobotHWGroup` sounds good to me.
- Regarding `joint_name_filter`, this is just a configuration example. The `CombinedRobotHW` doesn't impose any constraint on the parameters that a particular RobotHW would use for its set up. The only parameter that is necessary is `type` for obvious reasons.
- About read() and write(), if that is the case maybe I should change the signature of read and write to:

``` cpp
read(ros::Time time, ros::Duration period)
write(ros::Time time, ros::Duration period)
```

toliver:
@bmagyar 
- I've just checked and the changes actually don't break the API. The constructor without parameters would only be necessary if you want to publish your MyRobotHW class as a plugin by including:

``` cpp
PLUGINLIB_EXPORT_CLASS( my_robot_hw::MyRobotHW, hardware_interface::RobotHW)
```

in your cpp file. But that is of course acceptable as it is a requirement of `pluginlib`.
- Thanks for the merging tips. I will rebase this PR (and also #220) to comply with the best practice.
- I can try to write a gazebo example. The problem is that gazebo loads plugins that inherit from `gazebo_ros_control::RobotHWSim`, so I should maybe write a `CombinedRobotHWSim` version of `CombinedRobotHW` in order to test it with gazebo.

combined_robot_hw/include/combined_robot_hw/combined_robot_hw.h
toliver: Thanks, I didn't spot that.

combined_robot_hw/include/combined_robot_hw/combined_robot_hw.h
```diff
@@ -0,0 +1,115 @@
...
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+//////////////////////////////////////////////////////////////////////////////
+
+#ifndef COMBINED_ROBOT_HARDWARE_COMBINED_ROBOT_HARDWARE_H
+#define COMBINED_ROBOT_HARDWARE_COMBINED_ROBOT_HARDWARE_H
```
davetcoleman: should HARDWARE be HW?

combined_robot_hw/include/combined_robot_hw/combined_robot_hw.h
```diff
@@ -0,0 +1,115 @@
...
+#include <typeinfo>
+#include <hardware_interface/internal/demangle_symbol.h>
+#include <hardware_interface/internal/interface_manager.h>
+#include <hardware_interface/hardware_interface.h>
+#include <hardware_interface/robot_hw.h>
+#include <pluginlib/class_loader.h>
+#include <ros/console.h>
+#include <ros/node_handle.h>
+
+namespace combined_robot_hardware
```
davetcoleman: hardware should be hw

combined_robot_hw/include/combined_robot_hw/combined_robot_hw.h
```diff
@@ -0,0 +1,115 @@
...
+protected:
+  ros::NodeHandle root_nh_;
+  ros::NodeHandle robot_hw_nh_;
+  pluginlib::ClassLoader<hardware_interface::RobotHW> robot_hw_loader_;
+  std::vector<boost::shared_ptr<hardware_interface::RobotHW> > robot_hw_list_;
+
+  virtual bool loadRobotHW(const std::string& name);
+
+  /** \brief Filters the start and stop lists so that they only contain the controllers and
+   * resources that correspond to the r_hw interface manager
```
davetcoleman: Could you better define what r_hw is referring to? Just "robot_hw"?

combined_robot_hw/src/combined_robot_hw.cpp
```diff
@@ -0,0 +1,238 @@
...
+    robot_hw_loader_("hardware_interface", "hardware_interface::RobotHW")
+  {}
+
+  bool CombinedRobotHW::init(ros::NodeHandle& root_nh, ros::NodeHandle &robot_hw_nh)
+  {
+    root_nh_ = root_nh;
+    robot_hw_nh_ = robot_hw_nh;
+
+    std::vector<std::string> robots;
+    if (!robot_hw_nh.getParam("robot_hardware", robots)) {return false;}
```
davetcoleman: Would be nice to have error messages when it `return false`

combined_robot_hw/src/combined_robot_hw.cpp
```diff
@@ -0,0 +1,238 @@
...
+    }
+  }
+
+  bool CombinedRobotHW::loadRobotHW(const std::string& name)
+  {
+    ROS_DEBUG("Will load robot HW '%s'", name.c_str());
+
+    ros::NodeHandle c_nh;
+    // Constructs the robot HW
+    try{
```
davetcoleman: roscpp style guide: "Braces, both open and close, go on their own lines (no "cuddled braces")."

combined_robot_hw_tests/CMakeLists.txt
```diff
@@ -0,0 +1,63 @@
+cmake_minimum_required(VERSION 2.8.3)
+project(combined_robot_hw_tests)
```
davetcoleman, line 2: Do the tests really need their own package? Can't they just be inside combined_robot_hw?

combined_robot_hw_tests/test_robot_hw_plugin.xml
```diff
@@ -0,0 +1,25 @@
+<library path="lib/libcombined_robot_hw_tests">
```
davetcoleman, line 1: Pretty awesome tests, good job!

davetcoleman:
The changes to interface_manager.h are very complex, maybe @adolfo-rt can review those?

toliver:
Most of the changes in interface_manager.h belong to PR #220. It's probably easier to review that one first.

I'm not sure if it's good practice to submit a PR that includes the commits of a previous unmerged PR, but for this PR I needed to build on that.

combined_robot_hw_tests/CMakeLists.txt
toliver, line 2: It's not strictly necessary, but the fact that we are exporting plugins to be used only in the tests (MyRobotHW1, MyRobotHW2...) might make the package interface look untidy, so I followed the example of `controller_manager_tests` that I assumed had been created for the same reason.

Do you think it's better to have everything in one package?

combined_robot_hw/include/combined_robot_hw/combined_robot_hw.h
toliver: Yes, I'll change that to robot_hw.

toliver:
I have fixed some of the comments and also changed read() and write() to 

``` cpp
read(const ros::Time& time, const ros::Duration& period)
write(const ros::Time& time, const ros::Duration& period)
```

bmagyar:
Building on a different PR is fine, but risky since that one needs to be cleared first!

toliver:
Reopen against kinetic-devel as #235

combined_robot_hw_tests/CMakeLists.txt
davetcoleman, line 2: I am still wavering on this, but I see your point about hiding the test plugins and agree with your choice

bmagyar, line 2: It is the same approach we have for `controller_manager_tests` already btw

piyush2307:
Is it possible that we can control different hardware interface with different frequency in control loop.

PR #239, opened 2016-05-25
Update rosinstall versions to jade-devel

efernandez:
@bmagyar Merging this would help to test https://github.com/ros-controls/ros_controllers/pull/211


efernandez:
The test/compilation passed, but it's actually using the rosinstall from upstream. I don't see any way to test this other than actually merge it.

If needed, I could try to create a branch that would point to this branch to pick the rosinstall.
@bmagyar Thoughts?

bmagyar:
I think I saw it on some other discussions that Travis only runs tests for upstream branches. I'm by far not the expert on this, so do as you see fit :)

bmagyar:
Btw the changes are obviously ok 👍  the rosinstall should still be correct no matter what Travis checks

PR #240, opened 2016-05-25
Update rosinstall versions to kinetic-devel

efernandez:
Same as https://github.com/ros-controls/ros_control/pull/239, but for `kinetic-devel`

FYI @bmagyar


bmagyar:
Muy bien 👍

PR #247, opened 2016-08-05
Fix the example in the comments in multi_interface_controller.h.

miguelprada:
This is a very silly update, but it might save some time to anyone that copies and pastes from that example. If you don't figure out what's wrong on the first glance, the compiler errors can easily lead you down the templates rabbit hole.


bmagyar:
Hey Miguel, 
:+1: for catching this! I'm going to quickly merge this...

PR #252, opened 2016-10-27
Add combined_robot_hw to metapackage & system figure

bmagyar:
Forgot to add when packages were introduced. This should add it to the wiki top index too.
The architecture file was provided by Toni Oliver from Shadow Robot, made for his ROScon 2016 talk.


efernandez:
:+1: 

Although in the future (not part of this PR) I'd like to:
1. Upgrade to format 2
2. Alphabetize the list of depends
3. Use `test_depend` for test-only depends (which does the right thing with format 2)

bmagyar:
Thanks for the note about format 2!
I'll update this PR to that then ;)

bmagyar:
Change of plan...I'll do that separately. Want to see this do well on the doc jobs first.
Merging...

PR #253, opened 2016-10-27
Remove boost from depends declaration to fix cmake warning

bmagyar:
This is to remove 2 cmake warnings for said packages.

`catkin_package() DEPENDS on 'boost' but neither 'boost_INCLUDE_DIRS' nor 'boost_LIBRARIES' is defined.`

### DIFF
--- a/controller_interface/CMakeLists.txt
+++ b/controller_interface/CMakeLists.txt
@@ -6,7 +6,6 @@ find_package(catkin REQUIRED COMPONENTS roscpp hardware_interface pluginlib)
 
 # Declare catkin package
 catkin_package(
-  DEPENDS boost
   CATKIN_DEPENDS roscpp hardware_interface pluginlib
   INCLUDE_DIRS include
   )
--- a/controller_manager/CMakeLists.txt
+++ b/controller_manager/CMakeLists.txt
@@ -4,11 +4,13 @@ project(controller_manager)
 # Load catkin and all dependencies required for this package
 find_package(catkin REQUIRED COMPONENTS controller_interface controller_manager_msgs hardware_interface realtime_tools pluginlib)
 
+find_package(Boost REQUIRED COMPONENTS thread)
+
 include_directories(include ${Boost_INCLUDE_DIR} ${catkin_INCLUDE_DIRS})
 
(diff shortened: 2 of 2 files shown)


controller_interface/CMakeLists.txt
```diff
@@ -6,7 +6,6 @@ find_package(catkin REQUIRED COMPONENTS roscpp hardware_interface pluginlib)
 
 # Declare catkin package
 catkin_package(
-  DEPENDS boost
```
efernandez, line 9: It should be `Boost` in the `CMakeLists.txt` and `<depend>boost</depend>` in `package.xml` (assuming `format="2"`, but it's similar for format 1).

efernandez: However, if `Boost_INCLUDE_DIRS` and `Boost_LIBRARIES` aren't used in the `CMakeLists.txt`, I guess it's because nothing really depends on `Boost` here, right? Indeed, `find_package(Boost REQUIRED)` is needed as well, if `Boost` is actually used.

efernandez: `controller_manager` uses `Boost` here: https://github.com/ros-controls/ros_control/blob/kinetic-devel/controller_manager/src/controller_manager.cpp#L117

but `controller_interface` does NOT use it at all.

So, you should remove it here (`controller_interface`) and fix it for `controller_manager`.

efernandez (changes requested):
Remove `Boost` from `controller_interface` is OK, but `controller_manager` needs it, and should be fix to depend on it properly.

bmagyar:
I've rebased on the most recent kinetic-devel & did the proposed changes except for the package xml depend declaration. Do we really need to depend on boost there?

efernandez (approved):

efernandez:
Regarding the dependency on boost, I think ROS 1.0 depends on Boost, so in practice nothing bad happens if you don't depend on it because it's always there.

However, some packages depend on it: https://github.com/ros/ros_comm/blob/kinetic-devel/utilities/message_filters/package.xml#L17

It's up to you. You can merge it yourself when you're done (or as it's now). :smiley:

PR #259, opened 2016-12-07
Add urdf compatibility header

### LINKED ISSUE #258: Add urdf compatibility header for joint_limits_interface
To support urdfdom across different versions there's a new urdf compatibility header added to urdf. 

It was added here: https://github.com/ros/robot_model/commit/f5a3b8048c9fb4767ebd20b0cd80531c81522017

I believe that including that header will solve the symbol problem. Note that you'll need to add a dependency on urdf for this additional include. 

I believe that this is the root problem of this issue: https://github.com/ros-simulation/gazebo_ros_pkgs/issues/513

bmagyar:
Fixes #258


bmagyar:
I've tested it with `gazebo_ros_pkgs` on Wily and Xenial.

PR #261, opened 2017-01-26
controller_manager: drop unused inclusion of tinyxml.h

rojkov:


bmagyar:
I've ran some tests, all seems to be fine. 
Thanks for pinpointing this and for the PR!

:+1:

PR #267, opened 2017-03-09
Fixed claim policy for actuator command interface

nlamprian:
I believe the actuator command interface should claim the resources.

PR #272, opened 2017-05-10
Fixed claim policy for actuator command interface

nlamprian:
#267


bmagyar:
There's a glitch with the first test, I'll fix that. 
The rest reveals broken tests.

mathias-luedtke:
[This test](https://github.com/ros-controls/ros_control/blob/kinetic-devel/hardware_interface/test/actuator_command_interface_test.cpp#L125) tests explicitly that the resources are not claimed.
Looks like this was intended.

bmagyar:
Yep. On the top of my head I cannot bring a proper example but it feels like some funky transmissions could share actuators.

PR #273, opened 2017-05-10
Remove jade, update maintainers

bmagyar:


.travis.yml
```diff
@@ -1,14 +1,15 @@
 # Travis Continuous Integration Configuration File For ROS Control Projects
 # Author: Dave Coleman
 sudo: required
-dist: trusty
+dist: xenial
```
mathias-luedtke: This does not work. Travis does not support `xenial`, it will silently fall back to `precise`, which does not ship with docker (see https://github.com/ros-industrial/industrial_ci/pull/168 as well)

mathias-luedtke (dismissed):

.travis.yml
bmagyar: You are right, my bad!

graiola (approved):
Merged

Issue #275, opened 2017-06-27
How to propagate actuator commands from actuator to joint space


jlack1987:
I'm working in on the `kinetic-devel`  branch, and need to have the transmissions propagate actuator commands from actuator to joint space, but am having trouble figuring out how to make this happen because no ActuatorToJoint interfaces are created when the transmissions are created. I know this is not typically something you need to do, but humor me.

I'm trying to do all of this without copy/pasting/modifying all of the work done by the `transmission_interface_loader`. I feel like there's gotta be a "built in" way to make this happen, but i'm struggling to see it. Any advice?

jlack1987:
The avenue I see that might be the way to go is to create my own `RequisiteProvider` implementations and in my implementation of the `registerTransmission` method i'll register the `ActuatorToJoint_*_Cmd` interface along with the `jnt_to_act_cmd` interface that already gets registered so I can go both ways with it. Thoughts?

bmagyar:
So if I get the question right: you'd like to start a flow of information in the opposite direction? Is this about encoder feedback?

Could you sketch it up with draw.io or a small snippet of code?

jlack1987:
Yeah you got it. I want to be able to go in the opposite direction. It's not about encoder feedback. We will be using it as a way of making the simulation look like an actual piece of hardware. 

So picture a SimRobotHw in a gazebo plugin that will read joint state information from the sim robot, convert to actuator space, then give that to another HardwareRobotHw implementation that converts that actuator state info to joint state info in the `read()` method. Then in the `write()` method, it'll convert any joint commands to actuator commands and send it back over to the SimRobotHw. So now the SimRobotHw has a bunch of actuator commands it needs to convert back into joint space so it can send those commands to the simulated robots joints.

Edit: This kinda looks like,

```c++
class SimRobotHw : public RobotHw
{
    read()
   {
       sim_jnts_to_ros_control_jnts();
       propegate_jnt_state_to_act_state();
       publish(); // publish actuator data to be received by HardwareRobotHw in its read() method
   }

    write()
   {
       subscribe(); // Receive actuator commands from HardwareRobotHw in its write() method
       propegate_joint_cmd_to_act_cmd();
       ros_cntrl_joints_to_sim_jnts();
   }

}
```

Then the other side looks like a typical implementation of a `RobotHw` as you would expect and doesn't do any reverse propegation.

jlack1987:
[...]

PR #276, opened 2017-06-28
Add inverse transmission interfaces to transmission interface loader

jlack1987:
The purpose of this is to help me with [this issue](https://github.com/ros-controls/ros_control/issues/275) in which I would like to propegate transmissions in the opposite direction. In my own code I have created a `BiDirectionalEffortJointInterfaceProvider` plugin which is similar to the `EffortJointInterfaceProvider` plugin that registers the additional `ActuatorToJointEffortInterface`. The problem i'm running into now is the `TransmissionLoaderData` doesn't have these as members, so they aren't constructed and I can't register them as you all have done with the `ForwardTransmissionInterfaces`.

I could of course copy/paste what you all have done with the `TransmissionInterfaceLoader` and reimplement it myself, but i'd much rather add this slight modification which should not effect anything else and gives folks the opportunity to implement their own `RequisiteProvider` to register the reverse interfaces while still leveraging all the automatic transmission registration you all have done with the `TransmissionInterfaceLoader`. Let me know what you think.

### DIFF
--- a/transmission_interface/include/transmission_interface/bidirectional_effort_joint_interface_provider.h
+++ b/transmission_interface/include/transmission_interface/bidirectional_effort_joint_interface_provider.h
@@ -1,5 +1,5 @@
 ///////////////////////////////////////////////////////////////////////////////
-// Copyright (C) 2013, PAL Robotics S.L.
+// Copyright (C) 2017, Houston Mechatronics Inc.
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions are met:
@@ -8,7 +8,7 @@
 //   * Redistributions in binary form must reproduce the above copyright
 //     notice, this list of conditions and the following disclaimer in the
 //     documentation and/or other materials provided with the distribution.
-//   * Neither the name of PAL Robotics S.L. nor the names of its
+//   * Neither the name of Houston Mechatronics Inc. nor the names of its
 //     contributors may be used to endorse or promote products derived from
 //     this software without specific prior written permission.
 //
@@ -35,15 +35,14 @@
 
 namespace transmission_interface
 {
-    class BiDirectionalEffortJointInterfaceProvider : public EffortJointInterfaceProvider
-    {
-    protected:
-      
-        bool registerTransmission(TransmissionLoaderData& loader_data, 
-                                  TransmissionHandleData& handle_data);
-    };
+class BiDirectionalEffortJointInterfaceProvider : public EffortJointInterfaceProvider
+{
+protected:
+  
+  bool registerTransmission(TransmissionLoaderData& loader_data, 
+                            TransmissionHandleData& handle_data);
+};
 
 } // namespace
-
 
 #endif // header guard
--- a/transmission_interface/include/transmission_interface/bidirectional_effort_joint_interface_provider.h
+++ b/transmission_interface/include/transmission_interface/bidirectional_effort_joint_interface_provider.h
@@ -1,0 +1,49 @@
+///////////////////////////////////////////////////////////////////////////////
+// Copyright (C) 2013, PAL Robotics S.L.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//   * Redistributions of source code must retain the above copyright notice,
+//     this list of conditions and the following disclaimer.
+//   * Redistributions in binary form must reproduce the above copyright
+//     notice, this list of conditions and the following disclaimer in the
+//     documentation and/or other materials provided with the distribution.
+//   * Neither the name of PAL Robotics S.L. nor the names of its
+//     contributors may be used to endorse or promote products derived from
+//     this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+//////////////////////////////////////////////////////////////////////////////
+
+#ifndef TRANSMISSION_INTERFACE_BIDIRECTIONAL_EFFORT_JOINT_INTERFACE_PROVIDER_H
+#define TRANSMISSION_INTERFACE_BIDIRECTIONAL_EFFORT_JOINT_INTERFACE_PROVIDER_H
+
+// ros_control
+#include <transmission_interface/transmission_info.h>
+#include <transmission_interface/transmission_interface_loader.h>
+#include <transmission_interface/effort_joint_interface_provider.h>
+
+namespace transmission_interface
+{
+    class BiDirectionalEffortJointInterfaceProvider : public EffortJointInterfaceProvider
+    {
+    protected:
+      
+        bool registerTransmission(TransmissionLoaderData& loader_data, 
+                                  TransmissionHandleData& handle_data);
+    };
+
+} // namespace
+
+
+#endif // header guard
--- a/transmission_interface/include/transmission_interface/bidirectional_position_joint_interface_provider.h
+++ b/transmission_interface/include/transmission_interface/bidirectional_position_joint_interface_provider.h
@@ -1,0 +1,48 @@
+///////////////////////////////////////////////////////////////////////////////
+// Copyright (C) 2013, PAL Robotics S.L.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//   * Redistributions of source code must retain the above copyright notice,
+//     this list of conditions and the following disclaimer.
+//   * Redistributions in binary form must reproduce the above copyright
+//     notice, this list of conditions and the following disclaimer in the
+//     documentation and/or other materials provided with the distribution.
+//   * Neither the name of PAL Robotics S.L. nor the names of its
+//     contributors may be used to endorse or promote products derived from
+//     this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+//////////////////////////////////////////////////////////////////////////////
+
+#ifndef TRANSMISSION_INTERFACE_BIDIRECTIONAL_POSITION_JOINT_INTERFACE_PROVIDER_H
+#define TRANSMISSION_INTERFACE_BIDIRECTIONAL_POSITION_JOINT_INTERFACE_PROVIDER_H
+
+// ros_control
+#include <transmission_interface/transmission_info.h>
+#include <transmission_interface/transmission_interface_loader.h>
+#include <transmission_interface/position_joint_interface_provider.h>
+
+namespace transmission_interface
+{
+    class BiDirectionalPositionJointInterfaceProvider : public PositionJointInterfaceProvider
+    {
+    protected:
+
+        bool registerTransmission(TransmissionLoaderData& loader_data, 
+                                  TransmissionHandleData& handle_data);
+    };
+
+} // namespace
+
+#endif // header guard
--- a/transmission_interface/include/transmission_interface/bidirectional_position_joint_interface_provider.h
+++ b/transmission_interface/include/transmission_interface/bidirectional_position_joint_interface_provider.h
@@ -1,5 +1,5 @@
 ///////////////////////////////////////////////////////////////////////////////
-// Copyright (C) 2013, PAL Robotics S.L.
+// Copyright (C) 2017, Houston Mechatronics Inc.
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions are met:
@@ -8,7 +8,7 @@
 //   * Redistributions in binary form must reproduce the above copyright
 //     notice, this list of conditions and the following disclaimer in the
 //     documentation and/or other materials provided with the distribution.
-//   * Neither the name of PAL Robotics S.L. nor the names of its
+//   * Neither the name of Houston Mechatronics Inc. nor the names of its
 //     contributors may be used to endorse or promote products derived from
 //     this software without specific prior written permission.
 //
@@ -35,13 +35,13 @@
 
 namespace transmission_interface
 {
-    class BiDirectionalPositionJointInterfaceProvider : public PositionJointInterfaceProvider
-    {
-    protected:
+class BiDirectionalPositionJointInterfaceProvider : public PositionJointInterfaceProvider
+{
+protected:
 
-        bool registerTransmission(TransmissionLoaderData& loader_data, 
-                                  TransmissionHandleData& handle_data);
-    };
+  bool registerTransmission(TransmissionLoaderData& loader_data, 
+                            TransmissionHandleData& handle_data);
+};
 
 } // namespace
 
--- a/transmission_interface/include/transmission_interface/bidirectional_velocity_joint_interface_provider.h
+++ b/transmission_interface/include/transmission_interface/bidirectional_velocity_joint_interface_provider.h
@@ -1,0 +1,48 @@
+///////////////////////////////////////////////////////////////////////////////
+// Copyright (C) 2013, PAL Robotics S.L.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//   * Redistributions of source code must retain the above copyright notice,
+//     this list of conditions and the following disclaimer.
+//   * Redistributions in binary form must reproduce the above copyright
+//     notice, this list of conditions and the following disclaimer in the
+//     documentation and/or other materials provided with the distribution.
+//   * Neither the name of PAL Robotics S.L. nor the names of its
+//     contributors may be used to endorse or promote products derived from
+//     this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+//////////////////////////////////////////////////////////////////////////////
+
+#ifndef TRANSMISSION_INTERFACE_BIDIRECTIONAL_VELOCITY_JOINT_INTERFACE_PROVIDER_H
+#define TRANSMISSION_INTERFACE_BIDIRECTIONAL_VELOCITY_JOINT_INTERFACE_PROVIDER_H
+
+// ros_control
+#include <transmission_interface/transmission_info.h>
+#include <transmission_interface/transmission_interface_loader.h>
+#include <transmission_interface/velocity_joint_interface_provider.h>
+
+namespace transmission_interface
+{
+    class BiDirectionalVelocityJointInterfaceProvider : public VelocityJointInterfaceProvider
+    {
+    protected:
+
+        bool registerTransmission(TransmissionLoaderData& loader_data, 
+                                  TransmissionHandleData& handle_data);
+    };
+
+} // namespace
+
+#endif // header guard
--- a/transmission_interface/include/transmission_interface/bidirectional_velocity_joint_interface_provider.h
+++ b/transmission_interface/include/transmission_interface/bidirectional_velocity_joint_interface_provider.h
@@ -1,5 +1,5 @@
 ///////////////////////////////////////////////////////////////////////////////
-// Copyright (C) 2013, PAL Robotics S.L.
+// Copyright (C) 2017, Houston Mechatronics Inc.
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions are met:
@@ -8,7 +8,7 @@
 //   * Redistributions in binary form must reproduce the above copyright
 //     notice, this list of conditions and the following disclaimer in the
 //     documentation and/or other materials provided with the distribution.
-//   * Neither the name of PAL Robotics S.L. nor the names of its
+//   * Neither the name of Houston Mechatronics nor the names of its
 //     contributors may be used to endorse or promote products derived from
 //     this software without specific prior written permission.
 //
@@ -35,13 +35,13 @@
 
 namespace transmission_interface
 {
-    class BiDirectionalVelocityJointInterfaceProvider : public VelocityJointInterfaceProvider
-    {
-    protected:
+class BiDirectionalVelocityJointInterfaceProvider : public VelocityJointInterfaceProvider
+{
+protected:
 
-        bool registerTransmission(TransmissionLoaderData& loader_data, 
-                                  TransmissionHandleData& handle_data);
-    };
+  bool registerTransmission(TransmissionLoaderData& loader_data, 
+                            TransmissionHandleData& handle_data);
+};
 
 } // namespace
 
--- a/transmission_interface/src/bidirectional_effort_joint_interface_provider.cpp
+++ b/transmission_interface/src/bidirectional_effort_joint_interface_provider.cpp
@@ -1,5 +1,5 @@
 ///////////////////////////////////////////////////////////////////////////////
-// Copyright (C) 2013, PAL Robotics S.L.
+// Copyright (C) 2017, Houston Mechatronics Inc.
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions are met:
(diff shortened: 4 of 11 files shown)


jlack1987:
Really not trying to be pushy, but is there any chance I could get some developer feedback on this? This PR determines whether or not my group has to basically fork `ros_control` which I really don't want us to have to do. Thanks!

bmagyar:
It looks good to me at first glance although I feel it'd break ABI.

bmagyar (changes requested):
Please squash the commits.

bmagyar:
Could you add a simple test as well please?

jlack1987:
What is ABI? Yep I can try to come up with a test, although i'm not adding any functional code here, just adding the inverse interfaces to the `TransmissionLoaderData` object. And yeah i'll squash those commits, sorry about that.

jlack1987:
Commits squashed @bmagyar . Also i'm really not sure what unit test I could write since I just added a struct and not any moving parts that utilize the struct. Let me know if you have a specific case I should test.

bmagyar:
ABI stands for Application Binary Interface. It means binary compatibility beyond linking. 
It is something not nice do to within a ROS distro but it's not forbidden and I think this feature may be an important twist to add.

Have a look around the tests folder, you could have a test for a simple transmission that demonstrates propagating commands backwards. 
https://github.com/ros-controls/ros_control/blob/kinetic-devel/transmission_interface/test/transmission_interface_test.cpp#L241

jlack1987:
@bmagyar for me to add those unit tests I would need to include in this PR the 3 new `RequisiteProvider` implementations that register both the forward and reverse interfaces. I have written those and tested them on my end and can include them in this PR if you all are interested in that functionality. 

I couldn't come up with a good name for them so I called them, for example, a `BiDirectionalEffortJointInterfaceProvider` which is kinda a mouthful. Let me know if that name is ok and i'll include the 3 new interface providers and some tests for them.

bmagyar:
I'd like to see them, yes.
Don't worry about the long names, I bet all of us came across some Java code before ;)

What do you guys think? @ipa-mdl @graiola @efernandez

jlack1987:
Alright @bmagyar I added the extra requisite providers and tests for them all that pretty well follow the existing tests. 

I did my best to follow the style of the other files, but the styling is probably not perfect. Hoping you guys have some tool that'll autoreformat it all the way you like it.

efernandez:
LGTM, other than some style, the `const_cast` (that I honestly don't see how to avoid) and minor things I'd comment, but I don't want to be very picky yet (they're mostly things that `roslint` won't like if we used it).

bmagyar:
@jlack1987 I'm being pedantic now but I'd prefer to minimalize later patches in the core of ros_control. 
Could you please separate the code into 3 commits? Reviewing one mega-commit is hard to control.
* keeping the first commit as it was , only the core changes
* adding the interface
* adding tests

I feel that there is a lot of code repetition. Couldn't we get similar results by adding a flag to a template determining the interface direction?
(Of course tests are out of this question but not strictly, those are ok repeated but perhaps that could be parametrized the same way.)

efernandez:
@bmagyar I don't see how a template flag could be use here. It'd need a compilation time check, usually a type.

However, another option I see to avoid redundancy is the following:

1. Inherit the BiDirectional stuff from the non-bidirectional one.
2. In all virtual methods, call first the base class one.
3. Add the logic for the inverse transmissions.

If possible (likely not), in the non-bidirectional classes, the methods could be changed to use new helper methods that work on the transmission (probably templated).
This way we can call this helper methods from both the base and child class, if there's a bunch of code doing the same.

It'd be more or less like this:
``` c++
class BiDirectionalEffortJointInterfaceProvider : public EffortJointInterfaceProvider
{
  // SAME AS NOW
};
```
and (example for `registerTransmission`, other should be similar, I think)
``` c++
bool BiDirectionalEffortJointInterfaceProvider::registerTransmission(TransmissionLoaderData &loader_data,  TransmissionHandleData &handle_data)
{
  if (!EffortJointInterfaceProvider::registerTransmission(loader_data, handle_data))
  {
    return false;
  }

  if(!hasResource(handle_data.name, loader_data.inverse_transmission_interfaces.jnt_to_act_state))
  {
    if(!loader_data.robot_transmissions->get<transmission_interface::JointToActuatorStateInterface>())
    {
      loader_data.robot_transmissions->registerInterface(&loader_data.inverse_transmission_interfaces.jnt_to_act_state);
    }

    transmission_interface::JointToActuatorStateInterface &interface = *(loader_data.robot_transmissions->get<transmission_interface::JointToActuatorStateInterface>());

    // Update transmission interface
    transmission_interface::JointToActuatorStateHandle handle(handle_data.name, handle_data.transmission.get(), handle_data.act_state_data, handle_data.jnt_state_data);
    interface.registerHandle(handle);
  }

  if(!loader_data.robot_transmissions->get<transmission_interface::ActuatorToJointEffortInterface>())
  {
    loader_data.robot_transmissions->registerInterface(&loader_data.inverse_transmission_interfaces.act_to_jnt_eff_cmd);
  }

  transmission_interface::ActuatorToJointEffortInterface &reverse_interface = *(loader_data.robot_transmissions->get<transmission_interface::ActuatorToJointEffortInterface>());

  // Setup command interface
  transmission_interface::ActuatorToJointEffortHandle reverse_handle = transmission_interface::ActuatorToJointEffortHandle(handle_data.name, handle_data.transmission.get(), handle_data.act_cmd_data, handle_data.jnt_cmd_data);
  reverse_interface.registerHandle(reverse_handle);
  return true;
}
```

I don't know all the details and tiny changes in your PR @jlack1987 , mainly because the diff doesn't show anything (files are new, so it's difficult to compare them with the non-bidirectional classes). I hope this makes sense. :smiley:

efernandez:
Indeed, for the other methods (`updateJoinInterfaces`, `getJointCommandData`, `getActuatorCommandData`) it looks they don't need to be overloaded.

Maybe I'm missing something ?! 
@jlack1987 You did some style changes (IMO for good) in the code, so it's not easy to quickly see if they're the same or not.

bmagyar:
Thanks Enrique,  the template thing was only an idea and I think  it could
be done with a bool value and C++11 magic. The point was not that however
but to spark some new ideas and it seemed to have done that ;)

Good note on the style changes, it would be great to have those in the last
commit only.

efernandez:
Yeah, your comment took the right effect ;)
And now I think that regardless of the approach, this PR should definitely avoid duplicating that big amount of code. We all know how difficult would be to maintain things after that.

If there's a way to avoid creating another child class in the hierarchy, I'd actually prefere that.
It's certainly possible to use a Policy-based template, but I'd have to see where the class/plugin is actually defined. Essentially there it'd take the appropriate policy (we'd have new classes for the policies only, and update the non-bidirectional classes to take a template flag for bidirectionality). Maybe this is a little to complicated, but definitely possible and with no impact on runtime if done well.

jlack1987:
@efernandez @bmagyar Which direction would you like me to go with it? As @efernandez mentioned I can have the new bidirectional classes inherit most of the methods(except for `registerTransmissions`), but it looks like in a later comment you said you'd prefer to not add another level of inheritance. Let me know if that'd be an acceptable way to go for now.

efernandez:
@jlack1987 I'd be happy with another level of inheritance, as in my proposal.

I don't have a good picture of how we can do it w/o that, and I'm afraid it'd be over-complicated.

bmagyar:
Let's say that `ros_control` relies pretty heavily on inheritance, one already needs to analyse that to figure out how things work. I think it's ok to inherit, it is still the cleanest way.

coryleyhee:
Agreed. I'll throw that together and get the commits sorted out the way you requested. Thanks so much for the feedback!

jlack1987:
Agreed. I'll throw that together and get the commits sorted out the way you requested. Thanks so much for the feedback!

jlack1987:
Ok @bmagyar @efernandez I have reworked the inheritance as we discussed and have reworked the commits as requested.

transmission_interface/include/transmission_interface/bidirectional_effort_joint_interface_provider.h
```diff
@@ -0,0 +1,49 @@
...
+#define TRANSMISSION_INTERFACE_BIDIRECTIONAL_EFFORT_JOINT_INTERFACE_PROVIDER_H
+
+// ros_control
+#include <transmission_interface/transmission_info.h>
+#include <transmission_interface/transmission_interface_loader.h>
+#include <transmission_interface/effort_joint_interface_provider.h>
+
+namespace transmission_interface
+{
+    class BiDirectionalEffortJointInterfaceProvider : public EffortJointInterfaceProvider
```
efernandez: Follow the coding style from https://github.com/jlack1987/ros_control/blob/dfcb72375a3378b99e64971a46ef5f612f7dbbca/transmission_interface/include/transmission_interface/effort_joint_interface_provider.h#L41

That is:

1. Don't indent inside namespace
2. Indent only by two spaces

transmission_interface/include/transmission_interface/bidirectional_position_joint_interface_provider.h
```diff
@@ -0,0 +1,48 @@
...
+#define TRANSMISSION_INTERFACE_BIDIRECTIONAL_POSITION_JOINT_INTERFACE_PROVIDER_H
+
+// ros_control
+#include <transmission_interface/transmission_info.h>
+#include <transmission_interface/transmission_interface_loader.h>
+#include <transmission_interface/position_joint_interface_provider.h>
+
+namespace transmission_interface
+{
+    class BiDirectionalPositionJointInterfaceProvider : public PositionJointInterfaceProvider
```
efernandez: Same as first comment.

transmission_interface/include/transmission_interface/bidirectional_position_joint_interface_provider.h
```diff
@@ -0,0 +1,48 @@
+///////////////////////////////////////////////////////////////////////////////
+// Copyright (C) 2013, PAL Robotics S.L.
```
efernandez: Minor: update copyright (year, and your name / institution); this applies for this and all the new files.

transmission_interface/include/transmission_interface/bidirectional_velocity_joint_interface_provider.h
```diff
@@ -0,0 +1,48 @@
...
+#define TRANSMISSION_INTERFACE_BIDIRECTIONAL_VELOCITY_JOINT_INTERFACE_PROVIDER_H
+
+// ros_control
+#include <transmission_interface/transmission_info.h>
+#include <transmission_interface/transmission_interface_loader.h>
+#include <transmission_interface/velocity_joint_interface_provider.h>
+
+namespace transmission_interface
+{
+    class BiDirectionalVelocityJointInterfaceProvider : public VelocityJointInterfaceProvider
```
efernandez: Same as first comment.

transmission_interface/src/bidirectional_effort_joint_interface_provider.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+// Pluginlib
+#include <pluginlib/class_list_macros.h>
+
+// ros_control
+#include <transmission_interface/bidirectional_effort_joint_interface_provider.h>
+
+namespace transmission_interface
+{
+    bool BiDirectionalEffortJointInterfaceProvider::registerTransmission(TransmissionLoaderData &loader_data,
```
efernandez: Same as first comment (see https://github.com/jlack1987/ros_control/blob/dfcb72375a3378b99e64971a46ef5f612f7dbbca/transmission_interface/src/effort_joint_interface_provider.cpp#L38)

transmission_interface/src/bidirectional_effort_joint_interface_provider.cpp
```diff
@@ -0,0 +1,87 @@
...
+            TransmissionHandleData &handle_data)
+    {
+        // Setup joint state interface first (if not yet done)
```
efernandez: This is still repeated from https://github.com/jlack1987/ros_control/blob/dfcb72375a3378b99e64971a46ef5f612f7dbbca/transmission_interface/src/effort_joint_interface_provider.cpp#L124

Why don't you do this only?

``` cpp
if(!JointStateInterfaceProvider::registerTransmission(loader_data, handle_data))
{
  return false;
}
```

The code in this line is done inside that method, and even worse, there's code in that method you're repeating a few lines below here. I'd mark below the bits you should remove here, since they're already in the other/parent one.

transmission_interface/src/bidirectional_effort_joint_interface_provider.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+            transmission_interface::JointToActuatorStateInterface &interface = *(loader_data.robot_transmissions->get<transmission_interface::JointToActuatorStateInterface>());
+
+            // Update transmission interface
+            transmission_interface::JointToActuatorStateHandle handle(handle_data.name, handle_data.transmission.get(), handle_data.act_state_data, handle_data.jnt_state_data);
+            interface.registerHandle(handle);
+        }
+
+        // If command interface does not yet exist in the robot transmissions, add it and use internal data structures
+        if(!loader_data.robot_transmissions->get<transmission_interface::JointToActuatorEffortInterface>())
```
efernandez: This `if` block is here: https://github.com/jlack1987/ros_control/blob/dfcb72375a3378b99e64971a46ef5f612f7dbbca/transmission_interface/src/effort_joint_interface_provider.cpp#L130

transmission_interface/src/bidirectional_effort_joint_interface_provider.cpp
```diff
@@ -0,0 +1,87 @@
...
+        {
+            loader_data.robot_transmissions->registerInterface(&loader_data.transmission_interfaces.jnt_to_act_eff_cmd);
+        }
+
+        if(!loader_data.robot_transmissions->get<transmission_interface::ActuatorToJointEffortInterface>())
+        {
+            loader_data.robot_transmissions->registerInterface(&loader_data.inverse_transmission_interfaces.act_to_jnt_eff_cmd);
+        }
+
+        transmission_interface::JointToActuatorEffortInterface &interface = *(loader_data.robot_transmissions->get<transmission_interface::JointToActuatorEffortInterface>());
```
efernandez: This instruction is here: https://github.com/jlack1987/ros_control/blob/dfcb72375a3378b99e64971a46ef5f612f7dbbca/transmission_interface/src/effort_joint_interface_provider.cpp#L135

transmission_interface/src/bidirectional_effort_joint_interface_provider.cpp
```diff
@@ -0,0 +1,87 @@
...
+        transmission_interface::ActuatorToJointEffortInterface &reverse_interface = *(loader_data.robot_transmissions->get<transmission_interface::ActuatorToJointEffortInterface>());
+
+        // Setup command interface
+        transmission_interface::JointToActuatorEffortHandle handle(handle_data.name, handle_data.transmission.get(), handle_data.act_cmd_data, handle_data.jnt_cmd_data);
```
efernandez: These two are here: https://github.com/jlack1987/ros_control/blob/dfcb72375a3378b99e64971a46ef5f612f7dbbca/transmission_interface/src/effort_joint_interface_provider.cpp#L138

transmission_interface/src/bidirectional_effort_joint_interface_provider.cpp
```diff
@@ -0,0 +1,87 @@
...
+        interface.registerHandle(handle);
+
+        transmission_interface::ActuatorToJointEffortHandle reverse_handle = transmission_interface::ActuatorToJointEffortHandle(handle_data.name, handle_data.transmission.get(), handle_data.act_cmd_data, handle_data.jnt_cmd_data);
```
efernandez: This looks like greater than 120 columns, please follow the style in https://github.com/jlack1987/ros_control/blob/dfcb72375a3378b99e64971a46ef5f612f7dbbca/transmission_interface/src/effort_joint_interface_provider.cpp#L138

transmission_interface/src/bidirectional_position_joint_interface_provider.cpp
```diff
@@ -0,0 +1,95 @@
...
+// Pluginlib
+#include <pluginlib/class_list_macros.h>
+
+// ros_control
+#include <transmission_interface/bidirectional_position_joint_interface_provider.h>
+
+namespace transmission_interface
+{
+
+    bool BiDirectionalPositionJointInterfaceProvider::registerTransmission(TransmissionLoaderData& loader_data,
```
efernandez: The comments form the previous cpp file likely apply here too.

transmission_interface/src/bidirectional_velocity_joint_interface_provider.cpp
```diff
@@ -0,0 +1,89 @@
...
+// Pluginlib
+#include <pluginlib/class_list_macros.h>
+
+// ros_control
+#include <transmission_interface/bidirectional_velocity_joint_interface_provider.h>
+
+namespace transmission_interface
+{
+
+    bool BiDirectionalVelocityJointInterfaceProvider::registerTransmission(TransmissionLoaderData &loader_data, TransmissionHandleData &handle_data)
```
efernandez: The comments form the previous cpp file likely apply here too.

transmission_interface/test/transmission_interface_loader_test.cpp
```diff
@@ -390,6 +390,239 @@ TEST_F(TransmissionInterfaceLoaderTest, SuccessfulLoad)
...
+  JointHandle eff_jnt_handle_diff2 = eff_jnt_iface->getHandle(info_diff.joints_.back().name_);
+
+  // Propagate state forward
+  act_pos.assign(3,  50.0);
+  act_vel.assign(3, -50.0);
+  act_eff.assign(3,   1.0);
+
+  act_to_jnt_state->propagate();
+
+  EXPECT_NEAR( 1.5, pos_jnt_handle_red.getPosition(),   EPS);
```
efernandez: Remove extra space after `(` in this block of three.

transmission_interface/test/transmission_interface_loader_test.cpp
```diff
@@ -390,6 +390,239 @@ TEST_F(TransmissionInterfaceLoaderTest, SuccessfulLoad)
...
+  EXPECT_NEAR( 1.5, pos_jnt_handle_diff1.getPosition(), EPS);
+  EXPECT_NEAR( 0.5, pos_jnt_handle_diff2.getPosition(), EPS);
+
+  EXPECT_NEAR(-1.0, pos_jnt_handle_red.getVelocity(),   EPS);
+  EXPECT_NEAR(-1.0, pos_jnt_handle_diff1.getVelocity(), EPS);
+  EXPECT_NEAR( 0.0, pos_jnt_handle_diff2.getVelocity(), EPS);
+
+  EXPECT_NEAR( 50.0, pos_jnt_handle_red.getEffort(),     EPS);
```
efernandez: Align `EPS` arguments within this block of three.

transmission_interface/test/transmission_interface_loader_test.cpp
```diff
@@ -390,6 +390,239 @@ TEST_F(TransmissionInterfaceLoaderTest, SuccessfulLoad)
...
+  EXPECT_NEAR(1.0, act_eff_cmd_handle_diff1.getEffort(), EPS);
+  EXPECT_NEAR(1.0, act_eff_cmd_handle_diff2.getEffort(), EPS);
+
+  // Now propegate things in the reverse direction
+  RawJointDataMap* joint_data_map = &trans_iface_loader.getData()->raw_joint_data_map;
+  joint_data_map->operator[]("foo_joint").position = 1.5;
+  joint_data_map->operator[]("bar_joint").position = 1.5;
+  joint_data_map->operator[]("baz_joint").position = 0.5;
+
+  joint_data_map->operator[]("foo_joint").velocity = -1.;
```
efernandez: I'd prefer adding the `0` after the `.` for these six floating point literals, for readability and consistency with the rest of this file.

transmission_interface/test/transmission_interface_loader_test.cpp
```diff
@@ -390,6 +390,239 @@ TEST_F(TransmissionInterfaceLoaderTest, SuccessfulLoad)
...
+  joint_data_map->operator[]("bar_joint").velocity = -1.;
+  joint_data_map->operator[]("baz_joint").velocity = -2.;
+
+  joint_data_map->operator[]("foo_joint").effort = 5.;
+  joint_data_map->operator[]("bar_joint").effort = 10.;
+  joint_data_map->operator[]("baz_joint").effort = -5.;
+
+  jnt_to_act_state->propagate();
+
+  EXPECT_NEAR(50., act_state_handle_red.getPosition(),   EPS);
```
efernandez: Same applies here and below for the `.0`.

Certainly a pedantic comment, so feel free to ignore it. :smiley:

efernandez (changes requested):
You still need to re-use code from the parent classes properly.

Other than that, mainly coding style comments, and other minor ones.

Thanks for the rework and great unit test.

jlack1987:
Thanks for the feedback @efernandez! I definitely got lazy in there with my copy/pasting lol. Pushed a commit to address your comments.

transmission_interface/src/bidirectional_velocity_joint_interface_provider.cpp
```diff
@@ -0,0 +1,84 @@
...
+// ros_control
+#include <transmission_interface/bidirectional_velocity_joint_interface_provider.h>
+
+namespace transmission_interface
+{
+
+bool BiDirectionalVelocityJointInterfaceProvider::registerTransmission(TransmissionLoaderData &loader_data,
+                                                                       TransmissionHandleData &handle_data)
+{
+  if(!JointStateInterfaceProvider::registerTransmission(loader_data, handle_data))
```
efernandez: Remove this `if` block, it's already inside the parent: https://github.com/jlack1987/ros_control/blob/f4cd7642ccf2c702218b7025ddae4feb0585cee6/transmission_interface/src/velocity_joint_interface_provider.cpp#L127

Or maybe I'm missing something ?!

transmission_interface/src/bidirectional_velocity_joint_interface_provider.cpp
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  if(!VelocityJointInterfaceProvider::registerTransmission(loader_data,handle_data))
+  {
+    return false;
+  }
+
+  if(!hasResource(handle_data.name, loader_data.inverse_transmission_interfaces.jnt_to_act_state))
+  {
+    if(!loader_data.robot_transmissions->get<transmission_interface::JointToActuatorStateInterface>())
```
efernandez: You can remove the `transmission_interface::` namespace prefix

transmission_interface/src/bidirectional_velocity_joint_interface_provider.cpp
```diff
@@ -0,0 +1,84 @@
...
+    {
+      loader_data.robot_transmissions->registerInterface(&loader_data.inverse_transmission_interfaces.jnt_to_act_state);
+    }
+
+    transmission_interface::JointToActuatorStateInterface &interface = *(loader_data.robot_transmissions->get<transmission_interface::JointToActuatorStateInterface>());
```
efernandez: Same here and in many other places. You're already in the `transmission_interface::` namespace, so you can (and should) remove them.

transmission_interface/src/bidirectional_velocity_joint_interface_provider.cpp
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  transmission_interface::ActuatorToJointVelocityInterface &interface = *(loader_data.robot_transmissions->get<transmission_interface::ActuatorToJointVelocityInterface>());
+
+  transmission_interface::ActuatorToJointVelocityHandle handle(handle_data.name, 
+                                                               handle_data.transmission.get(), 
+                                                               handle_data.act_cmd_data, 
+                                                               handle_data.jnt_cmd_data);
+  interface.registerHandle(handle);
+
```
efernandez: pedantic: Remove empty line here for consistency with the other files.

transmission_interface/test/transmission_interface_loader_test.cpp
efernandez: Pending.

transmission_interface/test/transmission_interface_loader_test.cpp
```diff
@@ -390,6 +390,239 @@ TEST_F(TransmissionInterfaceLoaderTest, SuccessfulLoad)
...
+
+  EXPECT_NEAR(-50.,  act_state_handle_red.getVelocity(),   EPS);
+  EXPECT_NEAR(-150., act_state_handle_diff1.getVelocity(), EPS);
+  EXPECT_NEAR( 50.,  act_state_handle_diff2.getVelocity(), EPS);
+
+  EXPECT_NEAR(0.1,  act_state_handle_red.getEffort(),   EPS);
+  EXPECT_NEAR(0.05, act_state_handle_diff1.getEffort(), EPS);
+  EXPECT_NEAR(0.15, act_state_handle_diff2.getEffort(), EPS);
+
+  act_pos_cmd_handle_red.setCommand(3.);
```
efernandez: Pending.

transmission_interface/test/transmission_interface_loader_test.cpp
```diff
@@ -390,6 +390,239 @@ TEST_F(TransmissionInterfaceLoaderTest, SuccessfulLoad)
...
+  joint_data_map->operator[]("bar_joint").effort = 10.;
+  joint_data_map->operator[]("baz_joint").effort = -5.;
+
+  jnt_to_act_state->propagate();
+
+  EXPECT_NEAR(50.0, act_state_handle_red.getPosition(),   EPS);
+  EXPECT_NEAR(50.0, act_state_handle_diff1.getPosition(), EPS);
+  EXPECT_NEAR(50.0, act_state_handle_diff2.getPosition(), EPS);
+
+  EXPECT_NEAR(-50.,  act_state_handle_red.getVelocity(),   EPS);
```
efernandez: Pending.

transmission_interface/src/bidirectional_position_joint_interface_provider.cpp
```diff
@@ -0,0 +1,86 @@
...
+// ros_control
+#include <transmission_interface/bidirectional_position_joint_interface_provider.h>
+
+namespace transmission_interface
+{
+
+bool BiDirectionalPositionJointInterfaceProvider::registerTransmission(TransmissionLoaderData& loader_data,
+                                                                       TransmissionHandleData& handle_data)
+{
+  if(!JointStateInterfaceProvider::registerTransmission(loader_data, handle_data))
```
efernandez: Same comments as in the next cpp file

transmission_interface/src/bidirectional_effort_joint_interface_provider.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+// ros_control
+#include <transmission_interface/bidirectional_effort_joint_interface_provider.h>
+
+namespace transmission_interface
+{
+bool BiDirectionalEffortJointInterfaceProvider::registerTransmission(TransmissionLoaderData &loader_data,
+                                                                     TransmissionHandleData &handle_data)
+{
+  if(!JointStateInterfaceProvider::registerTransmission(loader_data, handle_data))
```
efernandez: Same comments as in the next cpp file

transmission_interface/include/transmission_interface/bidirectional_effort_joint_interface_provider.h
```diff
@@ -0,0 +1,49 @@
...
+class BiDirectionalEffortJointInterfaceProvider : public EffortJointInterfaceProvider
+{
+protected:
+  
+  bool registerTransmission(TransmissionLoaderData& loader_data, 
+                            TransmissionHandleData& handle_data);
+};
+
+} // namespace
+
```
efernandez, line 47: style: remove empty line (leave one, not two)

efernandez (changes requested):
I think there are still things that can be removed, just the first `if` block, I think.

There's also a few style comments pending and a few new ones.

After that it should be good to merge IMO.
@bmagyar Could you have a second look once @jlack1987 is done with my comments, please?

transmission_interface/src/bidirectional_velocity_joint_interface_provider.cpp
jlack1987: Nope you are correct!

transmission_interface/include/transmission_interface/bidirectional_effort_joint_interface_provider.h
jlack1987: Think I fixed this one. Not sure why it didn't register it.

efernandez: Looks good now.

transmission_interface/src/bidirectional_effort_joint_interface_provider.cpp
```diff
@@ -0,0 +1,79 @@
...
+  if(!EffortJointInterfaceProvider::registerTransmission(loader_data, handle_data))
+  {
+    return false;
+  }
+
+  if(!hasResource(handle_data.name, loader_data.inverse_transmission_interfaces.jnt_to_act_state))
+  {
+    if(!loader_data.robot_transmissions->get<JointToActuatorStateInterface>())
+    {
+        loader_data.robot_transmissions->registerInterface(&loader_data.inverse_transmission_interfaces.jnt_to_act_state);
```
efernandez: Too much indentation here.

transmission_interface/src/bidirectional_position_joint_interface_provider.cpp
```diff
@@ -0,0 +1,81 @@
...
+    if(!loader_data.robot_transmissions->get<JointToActuatorStateInterface>())
+    {
+      loader_data.robot_transmissions->registerInterface(&loader_data.inverse_transmission_interfaces.jnt_to_act_state);
+    }
+
+    JointToActuatorStateInterface &interface = *(loader_data.robot_transmissions->get<JointToActuatorStateInterface>());
+
+    // Update transmission interface
+    JointToActuatorStateHandle handle(handle_data.name, 
+                                                              handle_data.transmission.get(),
```
efernandez: And here

transmission_interface/src/bidirectional_velocity_joint_interface_provider.cpp
```diff
@@ -0,0 +1,79 @@
...
+    if(!loader_data.robot_transmissions->get<JointToActuatorStateInterface>())
+    {
+      loader_data.robot_transmissions->registerInterface(&loader_data.inverse_transmission_interfaces.jnt_to_act_state);
+    }
+
+    JointToActuatorStateInterface &interface = *(loader_data.robot_transmissions->get<JointToActuatorStateInterface>());
+
+    // Update transmission interface
+    JointToActuatorStateHandle handle(handle_data.name, 
+                                                              handle_data.transmission.get(),
```
efernandez: Here  too.

efernandez (changes requested):
Review the indentation please. I found three places where it was wrong. Maybe there are a few others.

@jlack1987 Please confirm you addressed all my previous comments.

If @bmagyar doesn't have anything to add, I'd merge this tomorrow.

efernandez:
Not sure if @ipa-mdl could add something as well :question:

jlack1987:
Thanks for the feedback @efernandez, @bmagyar any suggestions?

efernandez (approved):
Thanks for following our review comments @jlack1987 

LGTM, switching on the first green light ;)
@bmagyar @ipa-mdl ?

bmagyar (approved):

bmagyar:
Thank you for sticking with the PR, I'm :+1: on it. 

Anyone anything to add @ipa-mdl @graiola @efernandez ? 
Unless there are any new comments I'll merge this same time tomorrow ;)

bmagyar:
Merging...

jlack1987:
@bmagyar do you know when a new debian will be built and these changes can be obtain via `apt`?

bmagyar:
I've made the release of `ros_control` into `kinetic` and `lunar`. Keep an eye on [the release discussions](https://discourse.ros.org/c/release). It's been some time since the last release if I remember well. You can even ask for a sync is it's urgent, that may make it faster.

PR #283, opened 2017-09-21
fix constness in ForceTorqueSensorHandle

### LINKED ISSUE #282: The constructor of ForceTorqueSensorHandle
Hello, when I am using the hardware_interface::ForceTorqueSensorHandle, I confuse the self-define constructor. We should not want to change the data in this class, but the parameter list is asked a non-const pointer. The constructor define as follow, 
```
ForceTorqueSensorHandle(const std::string& name,
                          const std::string& frame_id,
                          double* force,
                          double* torque)
```
I don't known that whether it is a bug or not? Thank you.

mathias-luedtke:
fixes #282 

I kept the non-const constructor for the sake of ABI compatibility, not sure if it is really needed.


bmagyar:
:+1:

bmagyar (approved):

PR #290, opened 2017-11-16
Use public virtual for Controller inheritance.

mikepurvis:
I can put up some tests around this if it's otherwise a welcome change, but I wanted to put forward the suggestion for discussion prior to getting deeper into it. The proposed modification allows controllers to be extended via multiple inheritance. In particular, this is relevant where a new hardware interface type is being added as part of the extension, since that requires inheriting from a differently-templated version of `MultiInterfaceController` than what might have been in the original controller's hierarchy.

Consider the following diagram:

```
    ControllerBase
           |
      Controller
           |
  MultiInterfaceController
       /      \
FooController  \
         \      \
        FooBarController
```

In this scenario, `FooController` contains virtual methods which are a hook for an extending controller to be able to hitch on additional functionality. We already do this today with DiffDriveController— [our version has a virtual `wheelSpeedLimiter` function](https://github.com/clearpathrobotics/ros_controllers/blob/6954281c05dffcaf4de8727b6138a3ed8e54df01/diff_drive_controller/include/diff_drive_controller/diff_drive_controller.h#L138-L139), and there are (internal) controllers which use that hook to supply additional wheel-level speed constraints to the controller in real time. For example, based on higher-level knowledge such as position in the map or satisfying other system constraints.

The new capability that I would like is to be able to add _additional hardware interfaces_ in `FooBarController`, and to be able to do so while maintaining the is-a relationship of inheritance, rather than having to fake it by having the "original" controller as a member and using pass-through methods ([example](https://gist.github.com/mikepurvis/070a4f163986da4513e7907576aa25b9)).

Based on my testing, making this change proposed here allows the multiple inheritance to succeed, at the cost of two additional pointers in each controller's vtable.

Further reading: https://stackoverflow.com/a/8696486/109517

### DIFF
--- a/controller_manager_tests/CMakeLists.txt
+++ b/controller_manager_tests/CMakeLists.txt
@@ -16,6 +16,7 @@ catkin_package(
 #common commands for building c++ executables and libraries
 add_library(${PROJECT_NAME}
   src/effort_test_controller.cpp
+  src/extensible_controllers.cpp
   src/my_dummy_controller.cpp
   src/my_robot_hw.cpp
   src/pos_eff_controller.cpp
--- a/controller_manager_tests/include/controller_manager_tests/extensible_controllers.h
+++ b/controller_manager_tests/include/controller_manager_tests/extensible_controllers.h
@@ -1,0 +1,38 @@
+#include <controller_interface/multi_interface_controller.h>
+#include <hardware_interface/joint_command_interface.h>
+#include <pluginlib/class_list_macros.h>
+
+namespace controller_manager_tests
+{
+
+/**
+ * This controller supplies an intentional extension point in the form of the virtual
+ * "helper" function that the update method calls.
+ */
+typedef controller_interface::MultiInterfaceController<hardware_interface::VelocityJointInterface> BaseControllerInterface;
+class ExtensibleController : public virtual BaseControllerInterface
+{
+public:
+  bool init(hardware_interface::RobotHW* robot_hw,
+            ros::NodeHandle& root_nh, ros::NodeHandle& controller_nh);
+  virtual int helper();
+  void update(const ros::Time&, const ros::Duration&);
+};
+
+/**
+ * The derived controller not only overrides the virtual helper method, it also adds an additional
+ * hardware interface, in order to demonstrate the flexbility of this mechanism.
+ */
+typedef controller_interface::MultiInterfaceController<
+    hardware_interface::VelocityJointInterface, hardware_interface::EffortJointInterface> DerivedControllerInterface;
+class DerivedController : public ExtensibleController, public DerivedControllerInterface
+{
+public:
+  bool initRequest(hardware_interface::RobotHW* hw, ros::NodeHandle& nh, ros::NodeHandle& pnh,
+      controller_interface::ControllerBase::ClaimedResources& cr);
+  bool init(hardware_interface::RobotHW* robot_hw,
+            ros::NodeHandle& root_nh, ros::NodeHandle& controller_nh);
+  virtual int helper();
+};
+
+}  // namespace controller_manager_tests
--- a/controller_manager_tests/include/controller_manager_tests/extensible_controllers.h
+++ b/controller_manager_tests/include/controller_manager_tests/extensible_controllers.h
@@ -30,7 +30,6 @@
 
 #include <controller_interface/multi_interface_controller.h>
 #include <hardware_interface/joint_command_interface.h>
-#include <pluginlib/class_list_macros.h>
 
 namespace controller_manager_tests
 {
--- a/controller_manager_tests/include/controller_manager_tests/extensible_controllers.h
+++ b/controller_manager_tests/include/controller_manager_tests/extensible_controllers.h
@@ -1,3 +1,33 @@
+///////////////////////////////////////////////////////////////////////////////
+// Copyright (C) 2017, Clearpath Robotics Inc.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//   * Redistributions of source code must retain the above copyright notice,
+//     this list of conditions and the following disclaimer.
+//   * Redistributions in binary form must reproduce the above copyright
+//     notice, this list of conditions and the following disclaimer in the
+//     documentation and/or other materials provided with the distribution.
+//   * Neither the name of hiDOF, Inc. nor the names of its
+//     contributors may be used to endorse or promote products derived from
+//     this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+//////////////////////////////////////////////////////////////////////////////
+
+#ifndef CONTROLLER_MANAGER_TESTS_EXTENSIBLE_CONTROLLERS_H
+#define CONTROLLER_MANAGER_TESTS_EXTENSIBLE_CONTROLLERS_H
+
 #include <controller_interface/multi_interface_controller.h>
 #include <hardware_interface/joint_command_interface.h>
 #include <pluginlib/class_list_macros.h>
@@ -36,3 +66,5 @@ public:
 };
 
 }  // namespace controller_manager_tests
+
+#endif  // CONTROLLER_MANAGER_TESTS_EXTENSIBLE_CONTROLLERS_H
--- a/controller_manager_tests/src/extensible_controllers.cpp
+++ b/controller_manager_tests/src/extensible_controllers.cpp
@@ -26,6 +26,7 @@
 //////////////////////////////////////////////////////////////////////////////
 
 #include <controller_manager_tests/extensible_controllers.h>
+#include <pluginlib/class_list_macros.h>
 
 using namespace controller_manager_tests;
 
--- a/controller_manager_tests/src/extensible_controllers.cpp
+++ b/controller_manager_tests/src/extensible_controllers.cpp
@@ -1,3 +1,29 @@
+///////////////////////////////////////////////////////////////////////////////
+// Copyright (C) 2017, Clearpath Robotics Inc.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//   * Redistributions of source code must retain the above copyright notice,
+//     this list of conditions and the following disclaimer.
+//   * Redistributions in binary form must reproduce the above copyright
+//     notice, this list of conditions and the following disclaimer in the
+//     documentation and/or other materials provided with the distribution.
+//   * Neither the name of hiDOF, Inc. nor the names of its
+//     contributors may be used to endorse or promote products derived from
+//     this software without specific prior written permission.
+//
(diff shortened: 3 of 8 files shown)


bmagyar:
Good old C++ and that diamond problem...

On my side, I welcome the idea. This is something we discussed a few times with @efernandez and @artivis a few years back. Tests would be very welcome :)

efernandez (approved):
Also LGTM.

Other than the additional pointers in each controller's vtable, is there any other drawback?
And how would those additional pointers impact performance?

bmagyar:
It'd only be a slight performance drop of having to make a jump to resolve to the right function implementation.

This is a dry description of the thing: https://www.cprogramming.com/tutorial/virtual_inheritance.html

But I quite like the drawing here: https://stackoverflow.com/a/28521423 (although it's of virtual functions, it works the same way for class-level virtual inheritance)

efernandez:
My questions are related with the overhead introduced, as commented here: http://www.drdobbs.com/cpp/multiple-inheritance-considered-useful/184402074?pgno=2

Likely a small penalty, but that actually depends on how many times the virtual methods are being called.
I simply want to make clear we all agree that penalty is neglectable and won't impact significantly existing code. I've already approved it because I think it's neglectable.

I personally don't know offhand which methods will be called heavily, but I think at least the `update` mehod in `ControllerBase` will: https://github.com/mikepurvis/ros_control/blob/dde4d1a5ae7f6352ee07be264ae5ee127c2ee338/controller_interface/include/controller_interface/controller_base.h#L74, and it'll pay the penalty of the two pointers, right? (http://www.drdobbs.com/cpp/multiple-inheritance-considered-useful/184402074)

controller_manager_tests/CMakeLists.txt
```diff
@@ -15,18 +15,13 @@ catkin_package(
...
-  include/controller_manager_tests/vel_eff_controller.h
+  src/extensible_controllers.cpp
+  src/my_dummy_controller.cpp
+  src/my_robot_hw.cpp
   src/pos_eff_controller.cpp
-  include/controller_manager_tests/pos_eff_controller.h
   src/pos_eff_opt_controller.cpp
-  include/controller_manager_tests/pos_eff_opt_controller.h
-  src/my_dummy_controller.cpp
-  include/controller_manager_tests/my_dummy_controller.h
```
mikepurvis: Separate PR for this change in #293.

controller_manager_tests/src/extensible_controllers.cpp
```diff
@@ -0,0 +1,59 @@
...
+  // TODO: Publish this return value so that the test can validate it?
+  int foo = helper();
+}
+
+
+bool DerivedController::initRequest(hardware_interface::RobotHW* hw, ros::NodeHandle& nh, ros::NodeHandle& pnh,
+    controller_interface::ControllerBase::ClaimedResources& cr)
+{
+  return DerivedControllerInterface::initRequest(hw, nh, pnh, cr);
+}
```
mikepurvis, line 60: Having to pass through this call is my least favourite aspect of the whole business, but I can't really see a good way to get around it. It's not sufficient to just `using DerivedControllerInterface::initRequest;`, since that doesn't count as an actual override, so the main other alternative would be introducing some new template that would supply this function and take the things you want to extend as its arguments. I'm just not sure the savings in code is worth further obscuring what's really going on in here.

mikepurvis: (Of course, this approach of calling the other `init` method from `DerivedController::init` only works because `DerivedControllerInterface` is a superset of the original controller's interface. If that were not the case, potentially this method would have to call multiple `initRequest` methods, another argument against trying to overly standardize it...)

mikepurvis:
Okay, there's now a test and trivial demonstration controller here to show what this looks like. I think the main thing is figuring out from a design/architectural point of view if this is a worthwhile thing to try to support, or if it really would be better to insist on doing extension by composition.

----

Here's an example of achieving the same effect via composition (dropping these files into the PR changes allows the new tests to still pass):

https://gist.github.com/mikepurvis/070a4f163986da4513e7907576aa25b9

Pluses: IMO easier to reason about (it's a controller delegating part of its operation to a sub-controller), template shenanigans are minimized, no change required to core ros_control packages.

Minuses: A non-toy implementation would have to pass through a bunch more methods (starting, stopping, state_), and requiring the other class to provide the helper method override is awkward and friend-sharing data back and forth across that boundary could get messy.

Thoughts?

controller_manager_tests/CMakeLists.txt
bmagyar: Merged, please rebase this one without that commit.

controller_manager_tests/test/cm_test.cpp
```diff
@@ -497,6 +496,84 @@ TEST(CMTests, listControllers)
...
+    ASSERT_TRUE(list_client.call(srv));
+    ASSERT_EQ(srv.response.controller.size(), 2);
+    ControllerState state = srv.response.controller[0].name == "extensible_controller" ?
+        srv.response.controller[0] : srv.response.controller[1];
+    EXPECT_EQ("extensible_controller", state.name);
+    EXPECT_EQ("running", state.state);
+    ASSERT_EQ(1, state.claimed_resources.size());
+  }
+
+  // Start the the derived controller instead.
```
efernandez: `the the` -> `the`

controller_manager_tests/test/cm_test.cpp
```diff
@@ -497,6 +496,84 @@ TEST(CMTests, listControllers)
...
+    srv.request.strictness = srv.request.STRICT;
+    ASSERT_TRUE(switch_client.call(srv));
+    ASSERT_TRUE(srv.response.ok);
+  }
+
+  // Validate the controller is running and claimed one resource.
+  {
+    ListControllers srv;
+    ASSERT_TRUE(list_client.call(srv));
+    ASSERT_EQ(srv.response.controller.size(), 2);
```
efernandez: pedantic: `ASSERT_EQ(2, srv.response.controller.size());` expected value should go first (same happens a little below in this file)

controller_manager_tests/test/cm_test.cpp
```diff
@@ -497,6 +496,84 @@ TEST(CMTests, listControllers)
...
+    ASSERT_TRUE(switch_client.call(srv));
+    ASSERT_TRUE(srv.response.ok);
+  }
+
+  // Validate the controller is running and claimed two resources.
+  {
+    ListControllers srv;
+    ASSERT_TRUE(list_client.call(srv));
+    ASSERT_EQ(srv.response.controller.size(), 2);
+    ControllerState state = srv.response.controller[0].name == "derived_controller" ?
```
efernandez: Or `ControllerState state = srv.response.controller[srv.response.controller[0].name == "derived_controller" ? 0 : 1];`

controller_manager_tests/include/controller_manager_tests/extensible_controllers.h
```diff
@@ -0,0 +1,70 @@
...
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+//////////////////////////////////////////////////////////////////////////////
+
+#ifndef CONTROLLER_MANAGER_TESTS_EXTENSIBLE_CONTROLLERS_H
+#define CONTROLLER_MANAGER_TESTS_EXTENSIBLE_CONTROLLERS_H
+
+#include <controller_interface/multi_interface_controller.h>
+#include <hardware_interface/joint_command_interface.h>
+#include <pluginlib/class_list_macros.h>
```
efernandez: Move to the .cpp file?

controller_manager_tests/test/cm_test.cpp
mikepurvis: I know. I was torn between doing it right vs. matching the rest of this file.

controller_manager_tests/test/cm_test.cpp
mikepurvis: Makes sense, changed.

controller_manager_tests/include/controller_manager_tests/extensible_controllers.h
mikepurvis: Yes, done.

controller_manager_tests/CMakeLists.txt
mikepurvis: Done.

controller_manager_tests/src/extensible_controllers.cpp
mikepurvis: Follow up: It actually doesn't work to call both `initRequest` methods, since `init` is virtual, so they end up both calling the final class's one, rather than ones which correspond to their own classes. Given this, it's possible that indeed a universal `ExtensionInterfaceController` template (like `MultiInterfaceController`) could be developed, where you list only your new interface and all controllers you're extending, and it supplies an `initRequest` implementation which takes care of calling in order the init methods for the controllers you're extending.

Anyway, all of that would be API sugar on the basic principles already present in this PR.

mikepurvis:
Thoughts on this, reviewers?

bmagyar:
I'm generally positive about this PR :+1: 

I've just discovered that our [tutorials](http://wiki.ros.org/ros_control/Tutorials) page got rolled back for some reason... 

Could you please put together a short tutorial for the wiki? Feel free to use the test code as an example or other. I'll migrate the tutorials over from github again but I'll make sure to have a backup of yours there as well.

bmagyar (approved):

controller_manager_tests/src/extensible_controllers.cpp
bmagyar: `ros_control` already relies heavily on inheritance, I think it's better for this case than composition.

bmagyar:
@mikepurvis Please put together a short tutorial here on this thread. The test code or something simpler with a bit of explanation will do. I'll place it on the `ros_control` wiki and the ROS wiki as well.

Once we have that I'll merge ;)

mikepurvis:
Sounds good— sorry for the delay. There's a bit of a chicken and egg here, where the standard upstream ros_controllers are not really set up for being extended, much less multiple-extended. Which is fine, it just means any discussion/tutorials about this capability is a little bit abstract until the whole thing is better and more widely understood.

Either way, I can definitely write up something based on the extensible controllers tests which are part of this PR.

mikepurvis:
Okay, here's a first draft to kick the tires on:


Extensible Controllers
======================

As regular shared object files, each library of controllers is free to link whatever
external helper libraries it requires to do its job. However, there are two special cases
of modularizing controllers, which are discussed here.

Inheritance
-----------

If a controller exposes one or more virtual methods, then it is possible for an extending
controller to subclass that base controller, and override the virtual methods, providing
additional capabilities or constraints.

An example of this could be a generic mobile base controller (such as `diff_drive_controller`),
which is focused on accomplishing three basic tasks:

- Subscribing to `cmd_vel` to receive motion commands.
- Using the kinematic information from URDF to command wheels accordingly.
- Observing wheel states and publishing an `odom` topic.

In some cases, additional constraints (such as a "slow zone") may be possible to implement
within the local planner, or by placing a node in between `move_base` and the `ros_control`
node. However, there are some scenarios where correctly applying them requires knowledge
both of the kinematic intent (the Twist message) but also of the individual wheel speeds.
An example of this is where a low-level system such as a safety-rated laser scanner is
observing the wheels and will trip an e-stop if either wheel exceeds the safe threshold.

In this case, an attractive option is to define a subclass controller of the generic mobile
base controller, which can override a `wheelSpeedLimiter` or similar method. This way, the
functionality of the base controller may be reused without needing to refactor it into
a common library.

Multiple Inheritance
--------------------

For more complicated scenarios, it may be desirable to create a controller which is the
composite of multiple smaller controllers, but uses overridden virtual methods to allow
some element of co-operation between the lower level controllers, as defined by methods
in the composite.

Examples might include:

- Controlling a power-constrained platform, where only a certain maximum current may
  be drawn at once, across multiple actuators.
- Collaboration between a manipulator and mobile base, to get an end effector into an
  absolute position in world coordinates.
- Enforcement of soft safety constraints.

A difficulty arises when attempting this because of the C++ diamond problem, since both
controllers being extended will ultimately have ControllerBase as a base class:

```
        ControllerBase
        /           \
       /             \
FooController    BarController
       \             /
        \           /
      FooBarController
```

The ros_control framework resolves the diamond by declaring the base classes in the controller
chain as `public virtual`, but because of the diamond, _all_ methods which are virtual in
ControllerBase must be overridden in FooBarController, even if the only implementation is to call
the individual implementations in FooController and BarController. In particular, if MultiInterfaceController
is used for either or both of the controllers being extended, it will be necessary to override
the `initRequest` method and supply an implementation which either calls your init method,
or calls those supplied by the source controllers, in the preferred order.

As the multiple inheritance pattern for controllers becomes more known and used, it may
be possible to offer a template class for this use case, which would provide default
implementations of the standard functions, avoiding some of the boilerplate.

controller_manager_tests/src/extensible_controllers.cpp
mikepurvis: For any future readers, PR #301 is a general-purpose `CompositeController` which allows combining multiple controllers into one, and cleanly handles details like calling the subcontroller start and stop methods in order, etc.

Regarding the point made above about `init` being virtual— it actually _does_ work out to call the individual subcontrollers' `initRequest` methods, because `init` is a virtual member of the templated `Controller<>`/`MultiInterfaceController<>` classes, _not_ of `ControllerBase`. So if only `ControllerBase` is inherited with `public virtual`, then there will still be multiple instances of all the other classes _other than `ControllerBase`, which means exactly one initRequest/init function per subcontroller. This reality is what made the `CompositeController` template possible.

mikepurvis:
I'd like a temperature check on PR #301. If we can commit to merging that for Melodic, then the above proposed tutorial content should be reworked to use it.

Issue #291, opened 2017-11-21
Soft joint limits though parameter


miguelprada:
Helper functions to parse the the soft limits (useful to set up the `XXXXJointSoftLimitsInterface` variants) are only provided for the URDF model, but not for the parameter server. I would consider very helpful to configure such limits in a per-application basis using parameters instead of having to modify (or parametrize via xacro) the robot URDF model.

Is there any strong reason for this? Otherwise I can try to put together a quick PR with an additional `getSoftJointLimits(...)` method in `joint_limits_rosparam.h`.

miguelprada:
Addressed by #292.

PR #292, opened 2017-11-22
Add method to populate SoftJointLimits from ROS parameter server.

miguelprada:
Addresses #291.

If this is well accepted I will extend the existing tests.

### DIFF
--- a/joint_limits_interface/include/joint_limits_interface/joint_limits_rosparam.h
+++ b/joint_limits_interface/include/joint_limits_interface/joint_limits_rosparam.h
@@ -168,6 +168,78 @@ inline bool getJointLimits(const std::string& joint_name, const ros::NodeHandle&
   return true;
 }
 
+/**
+ * \brief Populate a SoftJointLimits instance from the ROS parameter server.
+ *
+ * It is assumed that the following parameter structure is followed on the provided NodeHandle. Unspecified parameters
+ * are simply not added to the joint limits specification.
+ * \code
+ * soft_joint_limits:
+ *   foo_joint:
+ *     soft_lower_limit: 0.0
+ *     soft_upper_limit: 1.0
+ *     k_position: 10.0
+ *     k_velocity: 10.0
+ * \endcode
+ *
+ * This specification is similar to the specification of the safety_controller tag in the URDF, adapted to the parameter server.
+ *
+ * \param[in] joint_name Name of joint whose limits are to be fetched.
+ * \param[in] nh NodeHandle where the joint limits are specified.
+ * \param[out] soft_limits Where soft joint limit data gets written into. Limits specified in the parameter server will overwrite
+ * existing values. Values in \p soft_limits not specified in the parameter server remain unchanged.
+ * \return True if a limits specification is found (ie. the \p soft_joint_limits/joint_name parameter exists in \p nh), false otherwise.
+ */
+inline bool getSoftJointLimits(const std::string& joint_name, const ros::NodeHandle& nh, SoftJointLimits& soft_limits)
+{
+  // Node handle scoped where the soft joint limits are defined
+  ros::NodeHandle soft_limits_nh;
+  try
+  {
+    const std::string soft_limits_namespace = "soft_joint_limits/" + joint_name;
+    if (!nh.hasParam(soft_limits_namespace))
+    {
+      ROS_DEBUG_STREAM("No soft joint limits specification found for joint '" << joint_name <<
+                       "' in the parameter server (namespace " << nh.getNamespace() + "/" + soft_limits_namespace << ").");
+      return false;
+    }
+    soft_limits_nh = ros::NodeHandle(nh, soft_limits_namespace);
+  }
+  catch(const ros::InvalidNameException& ex)
+  {
+    ROS_ERROR_STREAM(ex.what());
+    return false;
+  }
+
+  // Soft position limits
+  double soft_lower_limit;
+  if (soft_limits_nh.getParam("soft_lower_limit", soft_lower_limit))
+  {
+    soft_limits.min_position = soft_lower_limit;
+  }
+  double soft_upper_limit;
+  if (soft_limits_nh.getParam("soft_upper_limit", soft_upper_limit))
+  {
+    soft_limits.max_position = soft_upper_limit;
+  }
+
+  // Safety k_position
+  double k_position;
+  if (soft_limits_nh.getParam("k_position", k_position))
+  {
+    soft_limits.k_position = k_position;
+  }
+
+  // Safety k_velocity
+  double k_velocity;
+  if (soft_limits_nh.getParam("k_velocity", k_velocity))
+  {
+    soft_limits.k_velocity = k_velocity;
+  }
+
+  return true;
+}
+
 }
 
 #endif
--- a/joint_limits_interface/include/joint_limits_interface/joint_limits_rosparam.h
+++ b/joint_limits_interface/include/joint_limits_interface/joint_limits_rosparam.h
@@ -171,8 +171,8 @@ inline bool getJointLimits(const std::string& joint_name, const ros::NodeHandle&
 /**
  * \brief Populate a SoftJointLimits instance from the ROS parameter server.
  *
- * It is assumed that the following parameter structure is followed on the provided NodeHandle. Unspecified parameters
- * are simply not added to the joint limits specification.
+ * It is assumed that the following parameter structure is followed on the provided NodeHandle. Only completely specified soft
+ * joint limits specifications will be considered valid.
  * \code
  * joint_limits:
  *   foo_joint:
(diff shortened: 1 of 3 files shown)


joint_limits_interface/include/joint_limits_interface/joint_limits_rosparam.h
```diff
@@ -168,6 +168,78 @@ inline bool getJointLimits(const std::string& joint_name, const ros::NodeHandle&
...
+    soft_limits_nh = ros::NodeHandle(nh, soft_limits_namespace);
+  }
+  catch(const ros::InvalidNameException& ex)
+  {
+    ROS_ERROR_STREAM(ex.what());
+    return false;
+  }
+
+  // Soft position limits
+  double soft_lower_limit;
```
bmagyar: I'd prefer using `soft_limits_nh.hasParam(XYZ)` and calling `getParam` to copy directly onto the limits object. This saves us a line for each move and simplifies the code.

bmagyar (changes requested):

joint_limits_interface/include/joint_limits_interface/joint_limits_rosparam.h
miguelprada: Reasonable enough.

This makes me wonder whether `getParam` is guaranteed not to modify the value of the output variable if the parameter does not exist, which would further simplify the code. A quick test suggests that it doesn't, but since it's not mentioned in the documentation at all, I guess it's better not to rely on that behavior.

bmagyar:
@miguelprada , please go ahead with the tests.

mathias-luedtke:
I would prefer to upload the soft limits to `joint_limits` as well, enabled by `has_soft_limits`. This way all limits for each joint can be configured in one place.

miguelprada:
@ipa-mdl I was also in doubt about how to namespace the parameters. What do you think now?

I also set non-existent values to zero, and made `k_velocity` a requirement, as [here](https://github.com/ros/urdfdom/blob/ef7fb67e8bb698972c940dd00d26fb500d6a98f4/urdf_parser/src/joint.cpp#L215) (and as specified in the URDF [joint](http://wiki.ros.org/urdf/XML/joint) element).

joint_limits_interface/include/joint_limits_interface/joint_limits_rosparam.h
```diff
@@ -168,6 +168,85 @@ inline bool getJointLimits(const std::string& joint_name, const ros::NodeHandle&
...
+  }
+  else
+  {
+    ROS_DEBUG_STREAM("No soft joint limits specification found for joint '" << joint_name <<
+                     "' in the parameter server (namespace " << limits_nh.getNamespace() << ").");
+    return false;
+  }
+
+  // Safety k_position
+  if (limits_nh.hasParam("k_position"))
```
mathias-luedtke: This could be simplified to
```c++
limits_nh.param("k_position", soft_limits.k_position, 0.0);
```

joint_limits_interface/include/joint_limits_interface/joint_limits_rosparam.h
```diff
@@ -168,6 +168,85 @@ inline bool getJointLimits(const std::string& joint_name, const ros::NodeHandle&
...
+    limits_nh = ros::NodeHandle(nh, limits_namespace);
+  }
+  catch(const ros::InvalidNameException& ex)
+  {
+    ROS_ERROR_STREAM(ex.what());
+    return false;
+  }
+
+  // Required k_velocity
+  if (limits_nh.hasParam("k_velocity"))
```
mathias-luedtke:
```c++
if (!limits_nh.getParam("k_velocity", soft_limits.k_velocity)) 
{
  ROS_DEBUG_STREAM(...);
  return false;
}
```
?

mathias-luedtke:
I like this namespace layout!

But I don't like that many braces ;)
Feel free to adapt to my suggestions, or not.

PS: I really wonder why roscpp calls `hasParam` and `getParam` in `param()` as well..

miguelprada:
> I like this namespace layout!

> But I don't like that many braces ;)
> Feel free to adapt to my suggestions, or not.

Done!

> PS: I really wonder why roscpp calls hasParam and getParam in param() as well..

Maybe something related to my [prior comment](https://github.com/ros-controls/ros_control/pull/292#discussion_r152840597)?

miguelprada:
While writing the tests, it came to me that setting the optional values to their default value (i.e. `0.0`) if not found may not be the desired behavior. This makes it impossible to use the parameter server to override just some of the existing values in the URDF, such as [here](https://github.com/ros-controls/ros_control/blob/ee16564f3590fdd43faea3f74633824204215d58/joint_limits_interface/test/joint_limits_rosparam_test.cpp#L131).

If partial override is allowed, then it is also not clear to me whether the existence of a `k_velocity` should be required, or whether whichever values are found should be updated. And in the latter case, when the function should return `false` (or whether it should at all).

I think my preferred solution would be:
- Return `false` iff no `k_velocity` is found in parameter server and existing value is `0.0`
- Update only the values found in parameter server, leaving the rest unchanged

Didn't change anything yet. Will wait for your comments @bmagyar @ipa-mdl.

mathias-luedtke:
> This makes it impossible to use the parameter server to override just some of the existing values in the URDF, such as here.

I am not sure if it makes sense to change only one specific value of the soft limits, because they are coupled.

> Didn't change anything yet. Will wait for your comments

How do you like the following logic?

```c
bool ok = false;
if(nh.param("has_soft_limits", false)){
  if(!nh.hasParam("k_velocity") || soft_limits.k_velocity == 0.0){
    ROS_ERROR("k_velocity is required");
    return false;
  }

  ok = nh.getParam("k_velocity", soft_limits.k_velocity) || ok;
  ok = nh.getParam("k_position", soft_limits.k_position) || ok;
  ok = nh.getParam("soft_lower_limit", soft_limits.soft_lower_limit) || ok;
  ok = nh.getParam("soft_upper_limit", soft_limits.soft_upper_limit) || ok;

 if(!ok) {
    ROS_WARN("No soft limits found");
 }

}
return ok;

**Comment by miguelprada** [comment]
> I am not sure if it makes sense to change only one specific value of the soft limits, because they are coupled.

I think I agree.

> How do you like the following logic?

I find it a bit inconsistent. Whereas, in `getJointLimits`, `false` will only be returned if the limits namespace does not exist, your suggested logic would return `false` whenever any of the parameters is not found. Also, even though `false` would be returned, any existing values in the parameter server would be written to `soft_limits`.

How about?

```c++
if(nh.hasParam("k_position") && nh.hasParam("k_velocity") && nh.hasParam("soft_lower_limit") && nh.hasParam("soft_upper_limit"))
{
  nh.getParam("k_position", soft_limits.k_position);
  nh.getParam("k_velocity", soft_limits.k_velocity);
  nh.getParam("soft_lower_limit", soft_limits.min_position);
  nh.getParam("soft_upper_limit", soft_limits.max_position);
  return true;
}
return false;
```

This way, the values will be overriden and `true` will be returned if, and only if, all values are specified in the parameter server.

bmagyar:
Nice and consistent. Like it.

bmagyar:
I'm wondering if it's a good idea to print a warning when not all limits are specified. May be handy to save some people from gray hair but slightly verbose.

mathias-luedtke:
@miguelprada: Your version is fine for me. However, I still think we should add `has_soft_limits` to be consistent with the existing functions.

bmagyar (approved):

bmagyar:
As we are done with discussing naming and print details I've put my green light on it. 

@ipa-mdl please merge at your convenience.

miguelprada:
> @miguelprada: Your version is fine for me. However, I still think we should add has_soft_limits to be consistent with the existing functions.

A slight difference here is that, unlike `JointLimits` with `has_position_limits`, etc., `SoftJointLimits` doesn't have a `has_soft_limits` member.

I can see the need for it in `JointLimits`, since an instance can hold some but not other limit specifications. This is not the case with `SoftJointLimits`; it either has everything or nothing, and this can be already deduced from the return value of the function.

I won't refuse to add it if you really prefer to have it for consistence, though. If that's the case, would you rather add it to the `SoftJointLimits` as a member as well @ipa-mdl? And what would be your preferred semantics for the return value of `getSoftJointLimits` then?

mathias-luedtke:
`has_soft_limits` would just disable the parsing and make `getSoftJointLimits` return `false`.
I don't think there is a direct need for adding it to the structure.
We have used the `has_*_limits` flags disable limits for certain robots or environments, so this might be handy for soft limits as well. I don't insist on adding it.

mathias-luedtke (approved):
Current state LGTM.

miguelprada:
> We have used the has_*_limits flags disable limits for certain robots or environments, so this might be handy for soft limits as well. I don't insist on adding it.

I see your point and I think it's fair. Give me a while and I'll add it.

miguelprada:
Sorry for the dirty amend, but I had forgotten to update the tests in `e352562` and I feel like this PR already has way more commits than it should.

bmagyar:
Probably we should squash-merge it.

mathias-luedtke:
@miguelprada: if you agree, I will squash-merge. If not, just go ahead and rebase accordingly.

miguelprada:
Sure, go ahead @ipa-mdl.

PR #293, opened 2017-11-23
Drop includes from CMake library build.

mikepurvis:
Alphabetize, avoid listing unnecessary includes in `add_library` call.

### DIFF
--- a/controller_manager_tests/CMakeLists.txt
+++ b/controller_manager_tests/CMakeLists.txt
@@ -15,18 +15,12 @@ catkin_package(
 
 #common commands for building c++ executables and libraries
 add_library(${PROJECT_NAME}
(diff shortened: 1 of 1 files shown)


bmagyar (approved):

efernandez:
Maybe a little late, and I'm not really a fan of having the `include` files there, but usually the reason put them there is because IDEs like QTCreater uses the `CMakeLists.txt` file to discover the project files, and if you don't list the headers there it won't find them, or show them properly.

Just saying. :)

mikepurvis:
I guess that's fair, but that seems like an unfortunate limitation of the IDE. Couldn't it figure out this kind of thing based on gcc's `.d` files, or even just parsing the source themselves?

Issue #298, opened 2018-01-11
Switch MultiInterfaceController to be variadic template


mikepurvis:
This is possible in Kinetic as C++11 is now the target platform:

http://www.ros.org/reps/rep-0003.html#kinetic-kame-may-2016-may-2021

Would clean up the T1/T2/T3/T4 stuff in this file:

https://github.com/ros-controls/ros_control/blob/kinetic-devel/controller_interface/include/controller_interface/multi_interface_controller.h

mathias-luedtke:
> This is possible in Kinetic as C++11 is now the target platform:

I don't understand it this way. AFAIK everything must be compilable with C++11, but it is not the default target.

mikepurvis:
I think that was the status for Jade, which said that the target platform was:

> C++03 (C++11 features are not used, but code should compile when -std=c++11 is used)

Versus Kinetic:

> C++11 (GCC 4.9 on Linux, as it's the version that Debian Jessie ships with)

I think the real issue is that (I guess for compatibility reasons?) even the GCC 5 that Xenial comes with doesn't have C++11 as the default— the features are still behind a flag.

mathias-luedtke:
> C++11 (GCC 4.9 on Linux, as it's the version that Debian Jessie ships with)

This means that if you'd like to enforce c++11 support you can now do so for all  supported platforms, e.g. with a flag.

In addition there is the [rationale on c++](http://www.ros.org/reps/rep-0003.html#c) with recent fix https://github.com/ros-infrastructure/rep/pull/152 :

> Use of C++11/C++14 features and filesystem/networking/etc... TS's (Technical Specifications) is allowed if they are checked for at configure time and equivalent functionality can be provided without the extra compiler features. 

This will get changed for melodic (https://github.com/ros-infrastructure/rep/pull/139)

PR #299, opened 2018-01-11
Switch MultiInterfaceController to use variadic templates.

mikepurvis:
As proposed in #298.

C++11 is the supported platform in ROS Kinetic, so this should be a safe change: http://www.ros.org/reps/rep-0003.html#kinetic-kame-may-2016-may-2021

Besides general cleanup, this is of interest to better support composing multiple controllers, especially being able to iterate the interfaces used by the controllers being composed.

### DIFF
--- a/controller_interface/include/controller_interface/multi_interface_controller.h
+++ b/controller_interface/include/controller_interface/multi_interface_controller.h
@@ -30,10 +30,8 @@
 #ifndef CONTROLLER_INTERFACE_MULTI_INTERFACE_CONTROLLER_H
 #define CONTROLLER_INTERFACE_MULTI_INTERFACE_CONTROLLER_H
 
-#include <algorithm>
-#include <sstream>
-
 #include <controller_interface/controller_base.h>
+#include <controller_interface/internal/robothw_interfaces.h>
 #include <hardware_interface/internal/demangle_symbol.h>
 #include <hardware_interface/robot_hw.h>
 #include <hardware_interface/hardware_interface.h>
@@ -41,34 +39,6 @@
 
 namespace controller_interface
 {
-
-/** \cond HIDDEN_SYMBOLS */
-namespace internal
-{
-
-template <class T>
-bool hasInterface(hardware_interface::RobotHW* robot_hw);
-
-template <class T>
-void clearClaims(hardware_interface::RobotHW* robot_hw);
-
-template <class T>
-void extractInterfaceResources(hardware_interface::RobotHW* robot_hw_in,
-                               hardware_interface::RobotHW* robot_hw_out);
-
-template <class T>
-void populateClaimedResources(hardware_interface::RobotHW*      robot_hw,
-                              ControllerBase::ClaimedResources& claimed_resources);
-
-template <class T>
-std::string enumerateElements(const T& val,
-                              const std::string& delimiter = ", ",
-                              const std::string& prefix = "",
-                              const std::string& suffix = "");
-
-} // namespace
-/** \endcond */
-
 
 /**
  * \brief %Controller able to claim resources from multiple hardware interfaces.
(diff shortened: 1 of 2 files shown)


controller_interface/include/controller_interface/multi_interface_controller.h
```diff
@@ -172,25 +172,10 @@ std::string enumerateElements(const T& val,
  * };
  * \endcode
  *
- * \tparam T1 Hardware interface type.
+ * \tparam T... Hardware interface types.
  * This parameter is \e required.
```
mikepurvis, line 146: I can use a `static_assert` with `sizeof...` to enforce there always being >=1 interface specified, but I don't actually see a hard requirement anywhere for that being the case. Is it truly illegal to have a "controller" which has no interfaces, but still runs some code on the control thread in its update method?

controller_interface/include/controller_interface/multi_interface_controller.h
```diff
@@ -172,25 +172,10 @@ std::string enumerateElements(const T& val,
...
- *
- * \tparam T3 Hardware interface type.
- * This parameter is \e optional. Leave unspecified if controller only claims
- * resources from \e two hardware interfaces.
- *
- * \tparam T4 Hardware interface type.
- * This parameter is \e optional. Leave unspecified if controller only claims
- * resources from \e three hardware interfaces.
- *
- * \pre When specified, template parameters \c T1 to \c T4 must be different
```
mikepurvis, line 190: This could potentially also be validated using `static_assert`; it wasn't checked previously.

controller_manager_tests/CMakeLists.txt
```diff
@@ -1,6 +1,8 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(controller_manager_tests)
 
+set_directory_properties(PROPERTIES COMPILE_OPTIONS "-std=c++11")
```
mikepurvis: For now I've put this here, as it's the only package which actually _requires_ it. Users of controller_interface who don't use `multi_interface_controller.h` don't even need it.

That said, perhaps there should be a separate PR/discussion about doing this across the board.

mathias-luedtke:
I would postpone this to melodic. If I undestood it correctly, c++14 is the default for newer compilers.

controller_manager_tests/CMakeLists.txt
mathias-luedtke: But (existing!) users of `multi_interface_controller.h` need to set this (?).

mikepurvis: Yes, that's correct. So perhaps this can just be curiosity for now— we can build our system from a fork until the `melodic-devel` branch is created.

mathias-luedtke:
The header could branch on `BOOST_HAS_VARIADIC_TMPL` (since Boost 1.53), but without the std flags settings. So if you use a newer compiler, you can use the variadic templates. Otherwise the old header gets used. Again, this could be reverted later.

However, this would still be an ABI break, but it does only require a rebuild and no further patches..
@bmagyar: What do you think?

mikepurvis:
I'm not in that much of a rush. We can build from a fork for a few months until there's a `melodic-devel` branch to merge this into.

bmagyar:
Even though we treated ABI breaking pretty liberally in Kinetic, I think it is time to stop that. 

I agree with putting this in a `melodic-devel` branch once it is time for it and pretty happy that you guys can in the mean time build from your fork, it means the new code gets some mileage before it even makes it to our repo! Melodic has been added to `rosdistro` 7 days ago :) 

Feel free to leave this PR open as a reminder!

mikepurvis:
Note that as part of this change, it may actually be possible to make the original `Controller<>` template have this variadic behaviour, and simply alias `MultiInterfaceController` to it (or inherit and deprecate), all without breaking API at all— we would simply add two more init functions signatures like:

```
virtual bool init(T*... hw, ros::NodeHandle&, ros::NodeHandle&);
virtual bool init(T*... hw, ros::NodeHandle&);
```

Then the key is being able to call these correctly from the generic `initRequest`, which I think could be done by recursively packing a `std::tuple` with the correct hardware interface pointers. Will play with it and see what comes up.

But if it worked, the interface to the variadic `MultiInterfaceController<SingleInterface>` would be identical to that of `Controller<SingleInterface>`, with the added bonus of giving multi-interface users a nicer API than having to extract the interfaces themselves from a RobotHW pointer.

bmagyar:
@mikepurvis could you please re-target this PR against `melodic-devel`?

mathias-luedtke:
> could you please re-target this PR against melodic-devel?

@bmagyar: The target branch can be changed easily with the PR edit button (top right).

bmagyar:
Olala, thanks for the heads up! I wasn't aware __I__ could do it!

controller_manager_tests/CMakeLists.txt
mikepurvis: Since this is now targeting Melodic, I shall remove the compile options.

mikepurvis:
Would love to see some attention as well to #290, #301, and #302, but I'll let you merge this first and then take care of rebasing those afterward.

bmagyar:
Could you rebase this one please?

mikepurvis:
Done. I don't think the Kinetic CI jobs are going to pass, though— those should be disabled for the `melodic-devel` branch (and the Melodic job is failing due to dependency wait).

PR #301, opened 2018-02-06
CompositeController template for easy combining of controllers

mikepurvis:
This follows on #290 and #299, and demonstrates a generic template using the `public virtual` change to allow multiple controllers with polymorphic extension points to be easily combined into a monolithic co-operative controller.

Here for discussion for the moment; more docs and tests to come.

### DIFF
--- a/controller_interface/CMakeLists.txt
+++ b/controller_interface/CMakeLists.txt
@@ -14,3 +14,8 @@ catkin_package(
 install(DIRECTORY include/${PROJECT_NAME}/
   DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION})
 
+if(CATKIN_ENABLE_TESTING)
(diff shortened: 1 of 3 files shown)


bmagyar:
Could you rebase this one please?

mikepurvis:
Done.

mikepurvis:
Looks like we have a dependency-wait issue on Melodic. However, the fact that this passed on Kinetic probably indicates inadequate test coverage. :)

bmagyar:
Ok so this PR will remain off the first Melodic release until we have more tests, right?

mikepurvis:
It definitely works— we've been using it on a new robot internally for several weeks. And it's also new functionality, so safe to merge in that sense also. However, if you want to hold it to keep me honest, I won't be at all offended.

mikepurvis:
Closed in favour of #387.

PR #302, opened 2018-02-09
Single controller template

mikepurvis:
As hinted at in https://github.com/ros-controls/ros_control/pull/299#issuecomment-364166135, with C++11 it's now possible to achieve all features of `MultiInterfaceController` with a single `Controller` template that doesn't break the existing API. You basically get an `init()` function which includes pointers to all the hardware interfaces you specify in the template instantiation, eg:

```
class VelEffController : public controller_interface::Controller<VelocityJointInterface, EffortJointInterface>
{
public:
  bool init(VelocityJointInterface* v, EffortJointInterface* e, ros::NodeHandle &n) override;
}
```

With a single argument, this is the same as the legacy `Controller` usage. For compatibility with existing `MultiInterfaceController`, a deprecated adapter header is supplied which populates a reduced `RobotHW` as it worked before.

I'd propose this as a change for ROS Melodic, and removing the now deprecated `multi_interface_controller.h` in ROS N.

----

Note that the diff given here is pretty unhelpful. I'd suggest reviewing the new versions of the headers directly:

https://github.com/mikepurvis/ros_control/blob/single-controller-template/controller_interface/include/controller_interface/controller.h
https://github.com/mikepurvis/ros_control/blob/single-controller-template/controller_interface/include/controller_interface/multi_interface_controller.h

----

@adolfo-rt in particular I'm interested to hear from, as he was the original author of `MultiInterfaceController`, upon which this work builds heavily.

### DIFF
--- a/controller_interface/include/controller_interface/internal/robothw_interfaces.h
+++ b/controller_interface/include/controller_interface/internal/robothw_interfaces.h
@@ -1,0 +1,150 @@
+///////////////////////////////////////////////////////////////////////////////
+// Copyright (C) 2015, PAL Robotics S.L.
+// Copyright (C) 2017, Clearpath Robotics
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//   * Redistributions of source code must retain the above copyright notice,
+//     this list of conditions and the following disclaimer.
+//   * Redistributions in binary form must reproduce the above copyright
+//     notice, this list of conditions and the following disclaimer in the
+//     documentation and/or other materials provided with the distribution.
+//   * Neither the names of PAL Robotics S.L. nor the names of its
+//     contributors may be used to endorse or promote products derived from
+//     this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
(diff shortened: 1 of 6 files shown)


controller_interface/include/controller_interface/internal/robothw_interfaces.h
```diff
@@ -0,0 +1,151 @@
...
+                                     const std::string& prefix,
+                                     const std::string& suffix)
+{
+  std::string ret;
+  if (val.empty()) {return ret;}
+
+  const std::string sdp = suffix+delimiter+prefix;
+  std::stringstream ss;
+  ss << prefix;
+  std::copy(val.begin(), val.end(), std::ostream_iterator<class T::value_type>(ss, sdp.c_str()));
```
iwanders, line 60: Clang-7 complains about `class` here:
```
robothw_interfaces.h:60:68: error: typedef 'value_type' cannot be referenced with a class specifier
  std::copy(val.begin(), val.end(), std::ostream_iterator<class T::value_type>(ss, sdp.c_str()));
                                                                   ^
```
 using `typename` works without problems.

bmagyar:
What is the state of this? Is there something missing to merge?

controller_interface/include/controller_interface/internal/robothw_interfaces.h
jasonimercer: Also dies with Clang 6

mikepurvis:
Closed in favour of #387.

PR #306, opened 2018-03-20
migrate classloader headers

mathias-luedtke:
Just to get rid of the warnings.
Could be even cherry-picked for kinetic (after the next sync)


bmagyar:
Thank you!

PR #308, opened 2018-03-22
add urdf_geometry_parser to rosinstall file

mathias-luedtke:
Should be cherry-picked to kinetic as well.

### DIFF
--- a/ros_control.rosinstall
+++ b/ros_control.rosinstall
@@ -3,3 +3,4 @@
 - git: {local-name: realtime_tools, uri: 'https://github.com/ros-controls/realtime_tools.git', version: kinetic-devel}
 - git: {local-name: control_toolbox, uri: 'https://github.com/ros-controls/control_toolbox.git', version: kinetic-devel}
 - git: {local-name: control_msgs, uri: 'https://github.com/ros-controls/control_msgs.git', version: kinetic-devel}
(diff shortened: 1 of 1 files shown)


mathias-luedtke:
Not sure about adding `four_wheel_steering_msgs` in general.
We might move `four_wheel_steering_controller` into a separate repository though.

`ros_control` itself already  passes.
The other repos will hopefully pass after `xacro` was released.

bmagyar:
Yes, we should discuss a bettre strategy for splitting the repos and also for test setups.
Is there a way to specify per-job rosinstall files with industrial_ci?

bmagyar:
By the way, I've solved this issue for now by disabling the source-build of `ros_controllers` on the `ros_control` CI.

mathias-luedtke:
> Is there a way to specify per-job rosinstall files with industrial_ci?

You could  specify another file in `ROSINSTALL_FILENAME`.
In additon there is distro-based lookup: If  `$ROSINSTALL_FILENAME.$ROSDISTRO` (e.g. `ros_control.rosinstall.melodic`)  exists, it it gets used instead of `$ROSINSTALL_FILENAME`.

bmagyar:
The idea is that we could have a rosinstall with controllers and one
without.

Issue #316, opened 2018-04-09
Return policy mismatch between getJointLimits() and getSoftJointLimits()


yoshito-okada:
I am wondering why joint_limits_interface::getSoftJointLimits() returns false when no related parameters set although getJointLimits() returns true in the same situation.

I think returning true would be natural because these functions are intended to update the given limits as much as possible on the basis of parameters. So I would propose to remove [L225](https://github.com/ros-controls/ros_control/blob/16499ff46d3294ea32c0583be9da2b5c3bde5eb8/joint_limits_interface/include/joint_limits_interface/joint_limits_rosparam.h#L225) and to edit [L229](https://github.com/ros-controls/ros_control/blob/16499ff46d3294ea32c0583be9da2b5c3bde5eb8/joint_limits_interface/include/joint_limits_interface/joint_limits_rosparam.h#L229) to "return true" in joint_limits_rosparam.h.

miguelprada:
I see your point. However, I still think it would be nice to return false in case `has_soft_limits` is set and a complete soft limit specification can not be found. What do you think about:

```c++
inline bool getSoftJointLimits(const std::string& joint_name, const ros::NodeHandle& nh, SoftJointLimits& soft_limits)
{
//...
  // Override soft limits if complete specification is found
  bool has_soft_limits;
  if(limits_nh.getParam("has_soft_limits", has_soft_limits) && has_soft_limits)
  {
    if(!limits_nh.hasParam("k_position") || !limits_nh.hasParam("k_velocity") || !limits_nh.hasParam("soft_lower_limit") || !limits_nh.hasParam("soft_upper_limit"))
    {
      return false;
    }
    limits_nh.getParam("k_position", soft_limits.k_position);
    limits_nh.getParam("k_velocity", soft_limits.k_velocity);
    limits_nh.getParam("soft_lower_limit", soft_limits.min_position);
    limits_nh.getParam("soft_upper_limit", soft_limits.max_position);
  }

  return true;
}
// ...
```

It could be argued that nothing like this happens when parsing joint limits if, for instance `has_position_limits == true` and `min_position` or `max_position` are not set. Consistency with this behavior favors your proposed solution, @yoshito-okada. I still find it weird to return true if such an inconsistency in the parameters is found.

Thoughts, @ipa-mdl ?

yoshito-okada:
Almost agree.

[...]

PR #319, opened 2018-04-12
Modify getSoftJointLimits return value policy to match getJointLimits

miguelprada:
Related to #316


miguelprada:
Closing due to inactivity.

PR #324, opened 2018-04-17
Introduce shared_ptr typedefs

bmagyar:
Not the final version, more of a WIP.
These changes shouldn't introduce changes to existing API or ABI, only extending API with the typedefs.

### DIFF
--- a/combined_robot_hw/include/combined_robot_hw/combined_robot_hw.h
+++ b/combined_robot_hw/include/combined_robot_hw/combined_robot_hw.h
@@ -104,7 +104,7 @@ protected:
   ros::NodeHandle root_nh_;
   ros::NodeHandle robot_hw_nh_;
   pluginlib::ClassLoader<hardware_interface::RobotHW> robot_hw_loader_;
-  std::vector<boost::shared_ptr<hardware_interface::RobotHW> > robot_hw_list_;
+  std::vector<hardware_interface::RobotHWSharedPtr> robot_hw_list_;
 
   virtual bool loadRobotHW(const std::string& name);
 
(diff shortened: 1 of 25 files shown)


mathias-luedtke:
Looks more or less final to me. All public API is covered.
Only the transmission loader should get a typedef as well.

And `hwi_switch_test` should not use `boost::shared_ptr<Joint>` anymore.

bmagyar:
So...I've turned it from a changeset affecting 10 files to one that affects 25. Now it looks quite final :) 
There were a lot of local typedefs especially `TransmissionPtr` repeated many times. There is a single typedef I left in there that is now not being used by any of our code but remains part of the public API, I've put a deprecation warning in comment there. I think we should remove that for Melodic.

mathias-luedtke (approved):
Great job!

knorth55:
I have questions.
Is there a plan to backport this PR into `indigo-devel`?
Or is there any plan to merge this PR into `melodic-devel`?

mathias-luedtke:
> Or is there any plan to merge this PR into melodic-devel?

It's still work in progess: #331 

> Is there a plan to backport this PR into indigo-devel?

:+1: 
@bmagyar:  Do you agree? I can prepare the PR.

bmagyar:
Currently we are a bit stuck with this waiting for a pr of mine on
actionlib.

Yes it will go into melodic and kinetic too.

We could backport it into a last indigo release indeed. That'd ease
migration for all the folks still with indigo. If there's a volunteer
(wink-wink) I'd support it.

mathias-luedtke:
@bmagyar: can't we cherry-pick this to melodic without the C++14 adaption for now? Just to keep the API compatible.

bmagyar:
I agree. I'll try to fix it up tomorrow morning. If I remember well there's no C++14 there, only C++11 which if you enable it by hand can use in Kinetic too.

PR #331, opened 2018-05-02
WIP: C++11 changes

bmagyar:
Starts with cherry-picking the shared_ptr typedefs from kinetic

### DIFF
--- a/combined_robot_hw/src/combined_robot_hw.cpp
+++ b/combined_robot_hw/src/combined_robot_hw.cpp
@@ -62,7 +62,7 @@ namespace combined_robot_hw
                              const std::list<hardware_interface::ControllerInfo>& stop_list)
   {
     // Call the prepareSwitch method of the single RobotHW objects.
-    std::vector<boost::shared_ptr<hardware_interface::RobotHW> >::iterator robot_hw;
+    std::vector<hardware_interface::RobotHWSharedPtr>::iterator robot_hw;
     for (robot_hw = robot_hw_list_.begin(); robot_hw != robot_hw_list_.end(); ++robot_hw)
     {
       std::list<hardware_interface::ControllerInfo> filtered_start_list;
@@ -82,7 +82,7 @@ namespace combined_robot_hw
                         const std::list<hardware_interface::ControllerInfo>& stop_list)
   {
     // Call the doSwitch method of the single RobotHW objects.
-    std::vector<boost::shared_ptr<hardware_interface::RobotHW> >::iterator robot_hw;
+    std::vector<hardware_interface::RobotHWSharedPtr>::iterator robot_hw;
     for (robot_hw = robot_hw_list_.begin(); robot_hw != robot_hw_list_.end(); ++robot_hw)
     {
       std::list<hardware_interface::ControllerInfo> filtered_start_list;
@@ -117,7 +117,7 @@ namespace combined_robot_hw
       return false;
     }
 
-    boost::shared_ptr<hardware_interface::RobotHW> robot_hw;
+    hardware_interface::RobotHWSharedPtr robot_hw;
     std::string type;
     if (c_nh.getParam("type", type))
     {
@@ -187,7 +187,7 @@ namespace combined_robot_hw
   void CombinedRobotHW::read(const ros::Time& time, const ros::Duration& period)
   {
     // Call the read method of the single RobotHW objects.
-    std::vector<boost::shared_ptr<hardware_interface::RobotHW> >::iterator robot_hw;
+    std::vector<hardware_interface::RobotHWSharedPtr>::iterator robot_hw;
     for (robot_hw = robot_hw_list_.begin(); robot_hw != robot_hw_list_.end(); ++robot_hw)
     {
       (*robot_hw)->read(time, period);
@@ -198,7 +198,7 @@ namespace combined_robot_hw
   void CombinedRobotHW::write(const ros::Time& time, const ros::Duration& period)
   {
     // Call the write method of the single RobotHW objects.
-    std::vector<boost::shared_ptr<hardware_interface::RobotHW> >::iterator robot_hw;
+    std::vector<hardware_interface::RobotHWSharedPtr>::iterator robot_hw;
     for (robot_hw = robot_hw_list_.begin(); robot_hw != robot_hw_list_.end(); ++robot_hw)
     {
       (*robot_hw)->write(time, period);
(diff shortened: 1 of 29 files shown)


combined_robot_hw/src/combined_robot_hw.cpp
```diff
@@ -129,7 +129,7 @@ namespace combined_robot_hw
         {
           if (type == cur_types[i])
           {
-            robot_hw = robot_hw_loader_.createInstance(type);
+            robot_hw = robot_hw_loader_.createSharedInstance(type);
```
mathias-luedtke: `createUniqueInstance`, should auto-convert to `shared_ptr`.

controller_manager/include/controller_manager/controller_loader.h
```diff
@@ -74,7 +74,7 @@ class ControllerLoader : public ControllerLoaderInterface
 private:
   std::string package_;
   std::string base_class_;
-  boost::shared_ptr<pluginlib::ClassLoader<T> > controller_loader_;
+  std::shared_ptr<pluginlib::ClassLoader<T> > controller_loader_;
```
mathias-luedtke: `pluginlib::UniquePtr<T>`?

transmission_interface/include/transmission_interface/transmission_interface_loader.h
```diff
@@ -155,7 +155,7 @@ struct InverseTransmissionInterfaces
 
 struct TransmissionLoaderData
 {
-  typedef boost::shared_ptr<Transmission> TransmissionPtr;
+  typedef std::shared_ptr<Transmission> TransmissionPtr; // DEPRECATED and unused!
```
mathias-luedtke: remove?

combined_robot_hw/src/combined_robot_hw.cpp
bmagyar: `createSharedInstance` returns a `shared_ptr` directly

controller_manager/include/controller_manager/controller_loader.h
bmagyar: Good point! Adding...

transmission_interface/include/transmission_interface/transmission_interface_loader.h
bmagyar: Great catch, I've actually forgot about it

controller_manager/include/controller_manager/controller_loader.h
bmagyar: Hmm... I've changed my mind on this one. That typedef is only used as return type of `ClassLoader::createUniqueInstance()` while in this case it's merely a local variable of `ClassLoader<T>`

bmagyar:
Feel free to add more suggestions or commits to the PR!

combined_robot_hw/src/combined_robot_hw.cpp
mathias-luedtke: As travis confirmed, `createSharedInstance` does not (yet) exist. (https://github.com/ros/pluginlib/pull/67)

mathias-luedtke: https://github.com/ros/pluginlib/blob/b3d947b4ac92dadd27bbbdaaa3123b4c41aa6b07/include/pluginlib/class_loader_imp.hpp#L162

controller_manager/include/controller_manager/controller_loader.h
mathias-luedtke: My rationale was that we don't need shared ownership in this case.
In general `unique_ptr`  shoud be preferred over `shared_ptr`.

bmagyar: That's true. How about changing it to `std::unique_ptr` here?

combined_robot_hw/src/combined_robot_hw.cpp
bmagyar: True, I guess we can fly with that for now.

controller_manager/include/controller_manager/controller_loader.h
mathias-luedtke: ~~I would stick to `UniquePtr`, because this is used in the `pluginlib` code.~~

controller_manager/include/controller_manager/controller_loader_interface.h
```diff
@@ -50,7 +46,7 @@ class ControllerLoaderInterface
 {
 public:
   ControllerLoaderInterface(const std::string& name) : name_(name) { }
-  virtual boost::shared_ptr<controller_interface::ControllerBase> createInstance(const std::string& lookup_name) = 0;
+  virtual controller_interface::ControllerBaseSharedPtr createInstance(const std::string& lookup_name) = 0;
```
mathias-luedtke, line 49: `createInstance` -> `createSharedInstance` for consistency with `pluginlib`?

mathias-luedtke:
There are a lot of iterator-based for loops, should these be migrated to range-based for loops?

controller_manager/include/controller_manager/controller_loader.h
mathias-luedtke: You're right, `std::unique_ptr` is the way to go!

bmagyar:
In order to complete this first `realtime_tools` needs to be sorted out. There's an instance of `boost::shared_ptr` there that comes from `actionlib`. I'll keep working on this, probably adding a commit that undoes the realtime_tools typedef usage to keep the branch compiling while https://github.com/ros/actionlib/pull/106 gets merged & released.

mathias-luedtke:
> In order to complete this first realtime_tools needs to be sorted out.

I still don't get what needs to be sorted out here..

bmagyar:
Ooops, sorry. I'm mixing things in my head. `ros_controllers` needs to wait for `actionlib`, there are some typedefs that are supposed to be defined there but they weren't so they've been added to `realtime_tools` which I'd like to remove in Melodic.

This actually seems almost ready, just need to fix the current error.

mathias-luedtke:
>  just need to fix the current error.

fixed.

bmagyar:
Thank you! 
Should we go ahead and change the ControllerLoaderInterface API too? I had a brief check and I couldn't find instances of usage outside of the `ros_control` repo in standard packages. It's not likely anyone would implement their own `controller_manager`, right?

I'd like to merge this asap, don't want to keep it hanging for longer. We can add more C++11-ization later. There are still plenty of for loops with iterators that need shortening.

mathias-luedtke:
> Should we go ahead and change the ControllerLoaderInterface API too? 

What changes do you have in mind?

bmagyar:
Only the one you requested in the review.

mathias-luedtke:
> Only the one you requested in the review.

I see ;)
For now we should not change the API any further.

mathias-luedtke (approved):
3e1dd9dfce7dffefc5565cfbcdbc171c1f0bc833 should be merged sooner than later.

The pointer type change might break some downstream packages. It should get released ASAP, before too may of them migrated to melodic.

bmagyar:
I'll push a release now

PR #332, opened 2018-05-03
Add controller_group script that allows switching groups easily

yli-cpr:
The groups are read from ROS parameter "/controller_groups".

This is for the scenario that there can be multiple set of controllers that talk to same hardware interfaces but for different purposes. Without this script, one will have to stop all the production controllers and start a special set of controllers one by one. This script makes switching from a set to another much easier.

### DIFF
--- a/controller_manager/CMakeLists.txt
+++ b/controller_manager/CMakeLists.txt
@@ -46,7 +46,11 @@ install(DIRECTORY include/${PROJECT_NAME}/
 install(TARGETS ${PROJECT_NAME}
   LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION})
 
-install(PROGRAMS scripts/spawner scripts/unspawner scripts/controller_manager
+catkin_install_python(PROGRAMS
+  scripts/controller_group
+  scripts/controller_manager
+  scripts/spawner
+  scripts/unspawner
   DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION})
 
 catkin_python_setup()
--- a/controller_manager/scripts/controller_group
+++ b/controller_manager/scripts/controller_group
@@ -1,0 +1,135 @@
+#!/usr/bin/env python
+# Copyright (c) 2018, Clearpath Robotics, Inc.
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+#     * Redistributions of source code must retain the above copyright
+#       notice, this list of conditions and the following disclaimer.
+#     * Redistributions in binary form must reproduce the above copyright
+#       notice, this list of conditions and the following disclaimer in the
+#       documentation and/or other materials provided with the distribution.
+#     * Neither the name of the Willow Garage, Inc. nor the names of its
+#       contributors may be used to endorse or promote products derived from
+#       this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+#
+# Support controller groups so it is easy to spawn a set controllers and also switch
+# to any other group at run time.
+#
+# Author: Yong Li
+#
+import rospy
+from argparse import ArgumentParser
+from controller_manager import controller_manager_interface
+from controller_manager_msgs.srv import ListControllers, ListControllersRequest
+
+
+def _read_group(group):
+    ''''
+    Return 2 sets of controllers. The first set includes controllers required for the given group,
+    and the second set includes the controllers that should be stopped if the given group is chosen.
+    '''
+    groups = rospy.get_param('controller_groups', {})
+    desired_controllers = set(groups.get(group, ()))
+    if not desired_controllers:
+        rospy.logerr('There is no controller defined in group {}'.format(group))
+        return None, None
+    # Remove duplicates.
+    all_controllers = set()
+    for controllers in groups.values():
+        all_controllers.update(controllers)
+    return desired_controllers, all_controllers - desired_controllers
+
+
+def _get_loaded_controllers():
+    rospy.wait_for_service('controller_manager/list_controllers', timeout=5)
+    s = rospy.ServiceProxy('controller_manager/list_controllers', ListControllers)
+    resp = s(ListControllersRequest())
+    controllers = {}
+    for controller in resp.controller:
+        controllers[controller.name] = controller.state == 'running'
+        if controller.state not in ('running', 'stopped'):
+            rospy.logerr('Unknown controller state {}'.format(controller.state))
+    return controllers
+
+
+def _list_group(args):
+    groups = rospy.get_param('controller_groups', {})
+    print('Number of groups: {}'.format(len(groups)))
+    for group, controllers in sorted(groups.items()):
+        print('  {}'.format(group))
+        for controller in controllers:
+            print('      {}'.format(controller))
+
+
+def _spawn_group(args):
+    desired_controllers, _ = _read_group(args.group)
+    if not desired_controllers:
+        return
+    for controller in desired_controllers:
+        controller_manager_interface.load_controller(controller)
+    for controller in desired_controllers:
+        controller_manager_interface.start_controller(controller)
+
+
+def _switch_group(args):
+    desired_controllers, other_controllers = _read_group(args.group)
+    if not desired_controllers:
+        return
+    loaded_controllers = _get_loaded_controllers()
+
+    # Stop running controllers that we no longer need.
+    to_stop = []
+    for controller in other_controllers:
+        if loaded_controllers.get(controller, False):
+            to_stop.append(controller)
+
+    # Load and start every desired controller if needed.
+    to_start = []
+    for controller in desired_controllers:
+        if controller not in loaded_controllers:
+            if not controller_manager_interface.load_controller(controller):
+                # load_controller() already prints an error.
+                print('Cancel group switching')
+                return
+        if not loaded_controllers.get(controller, False):
+            to_start.append(controller)
+
+    controller_manager_interface.start_stop_controllers(start_controllers=to_start, stop_controllers=to_stop)
+
+
+def _main():
+    parser = ArgumentParser(description='Controller Group Utility')
+    subparsers = parser.add_subparsers(title='Commands', metavar='')
+
+    list_parser = subparsers.add_parser('list', help='list available groups')
+    list_parser.set_defaults(func=_list_group)
+
+    spawn_parser = subparsers.add_parser(
+        'spawn', help='load all controllers in all groups and start controllers in the specified group')
+    spawn_parser.add_argument('group', help='name of the group to start')
+    spawn_parser.set_defaults(func=_spawn_group)
+
+    switch_parser = subparsers.add_parser('switch', help='switch to the specified group')
+    switch_parser.add_argument('group', help='name of the group to switch to')
+    switch_parser.set_defaults(func=_switch_group)
+
+    args = parser.parse_args(rospy.myargv()[1:])
+    args.func(args)
+
+
+if __name__ == '__main__':
+    _main()
--- a/controller_manager/src/controller_manager/controller_manager_interface.py
+++ b/controller_manager/src/controller_manager/controller_manager_interface.py
@@ -75,37 +75,28 @@ def unload_controller(name):
         return False
 
 def start_controller(name):
-    return start_stop_controllers([name], True)
+    return start_stop_controllers(start_controllers=[name])
 
 def start_controllers(names):
-    return start_stop_controllers(names, True)
+    return start_stop_controllers(start_controllers=names)
 
 def stop_controller(name):
-    return start_stop_controllers([name], False)
+    return start_stop_controllers(stop_controllers=[name])
 
 def stop_controllers(names):
-    return start_stop_controllers(names, False)
+    return start_stop_controllers(stop_controllers=names)
 
-def start_stop_controllers(names, st):
+def start_stop_controllers(start_controllers=[], stop_controllers=[]):
     rospy.wait_for_service('controller_manager/switch_controller')
     s = rospy.ServiceProxy('controller_manager/switch_controller', SwitchController)
-    start = []
-    stop = []
     strictness = SwitchControllerRequest.STRICT
-    if st:
-        start = names
-    else:
-        stop = names
-    resp = s.call(SwitchControllerRequest(start, stop, strictness))
+    resp = s.call(SwitchControllerRequest(start_controllers, stop_controllers, strictness))
     if resp.ok == 1:
-        if st:
-            print "Started {} successfully".format(names)
-        else:
-            print "Stopped {} successfully".format(names)
+        if start_controllers:
+            print "Started {} successfully".format(start_controllers)
+        if stop_controllers:
+            print "Stopped {} successfully".format(stop_controllers)
         return True
     else:
-        if st:
-            print "Error when starting ", names
-        else:
-            print "Error when stopping ", names
+        print "Error when starting {} and stopping {}".format(start_controllers, stop_controllers)
         return False
--- a/controller_manager_tests/test/controller_manager_scripts.py
+++ b/controller_manager_tests/test/controller_manager_scripts.py
@@ -23,6 +23,10 @@ reload_response = 'Restore: False\nSuccessfully reloaded libraries\n'
 def run_cm(*args):
     return subprocess.check_output('rosrun controller_manager controller_manager ' + ' '.join(args), shell=True)
 
+# run controller_group in process
+def run_cg(*args):
+    return subprocess.check_output('rosrun controller_manager controller_group ' + ' '.join(args), shell=True)
+
 # helper function that return the actual and expected response
 def load_c(name):
     return run_cm('load', name), loaded_fmt % name
@@ -41,6 +45,7 @@ def spawn_c(name):
 
 def kill_c(name):
     return run_cm('kill', name), stopped_fmt % name + unloaded_fmt % name
+
 
 class TestUtils(unittest.TestCase):
     def test_scripts(self):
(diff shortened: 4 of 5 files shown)


controller_manager/scripts/controller_group
```diff
@@ -0,0 +1,96 @@
...
+        controllers = {}
+        for controller in resp.controller:
+            controllers[controller.name] = controller.state == 'running'
+            if controller.state not in ('running', 'stopped'):
+                rospy.logerr('Unknown controller state {}'.format(controller.state))
+        return controllers
+
+
+def _main():
+    rospy.init_node('controller_group', anonymous=True)
```
mikepurvis: Initializing a node may actually not be necessary, since you're only calling services and not interacting with topics at all.

controller_manager_tests/test/controller_manager_scripts.test
```diff
@@ -8,6 +8,13 @@
       type: controller_manager_tests/EffortTestController
       joints:
       - hiDOF_joint3
+    controller_groups:
+      group1:
+        - my_controller1
+        - my_controller3
+      group2:
+        - my_controller2
+        - my_controller3
```
mikepurvis, line 17: Should there be some kind of failure case here, like a group where two of the controllers try to claim the same resource? Does the new tool fail gracefully under those circumstances?

controller_manager/scripts/controller_group
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env python
+# Provides quick access to the services exposed by controller_manager_msgs
+import rospy
+from argparse import ArgumentParser
+from controller_manager import controller_manager_interface
+from controller_manager_msgs.srv import ListControllers, ListControllersRequest
+
+
+class Groups(object):
```
mikepurvis: Since this class has no state, IMO it's superfluous. Consider moving the methods up to module level (eg `list_groups`, `spawn_group`, etc).

controller_manager/scripts/controller_group
```diff
@@ -0,0 +1,96 @@
...
+    @staticmethod
+    def list(args):
+        groups = rospy.get_param('/controller_groups', {})
```
mikepurvis: IMO this is fine for now, but down the road it might be desirable to have this be configurable somehow, possibly an arg which allows passing the name of the group param?

controller_manager/scripts/controller_group
```diff
@@ -0,0 +1,96 @@
...
+        print('Number of groups: {}'.format(len(groups)))
+        for group, controllers in groups.iteritems():
```
mikepurvis: For py3 friendliness, prefer just `items`. For such a small list, the savings from the iterator are extremely small anyway.

mikepurvis (changes requested):
Looks great, mostly just piddly stuff.

controller_manager/scripts/controller_group
yli-cpr: Added for rospy.myargv() but seems not necessary.

controller_manager_tests/test/controller_manager_scripts.test
yli-cpr: I am trying a failure case. The script is not that smart to verify this error, but it will print errors.

controller_manager/scripts/controller_group
yli-cpr: sure

controller_manager/scripts/controller_group
yli-cpr: Sure. Also I need sorted() for python 2 to print stable output

controller_manager/scripts/controller_group
yli-cpr: Yeah. Also it can be a namespace? CombineRobotHW is loading `robot_hardware` from the node's namespace. I don't have much time available, so I would leave it for now.

controller_manager/scripts/controller_group
mikepurvis: That's a good point— this definitely does still need to work when run from a launch file.

controller_manager_tests/test/controller_manager_scripts.test
mikepurvis: The underlying SwitchController service call should return a `False`:

http://docs.ros.org/lunar/api/controller_manager_msgs/html/srv/SwitchController.html

So I think it's mostly just on the script here to detect that and bubble some kind of error message to the user.

controller_manager/scripts/controller_group
yli-cpr: do you mean init_node is still needed?

controller_manager/scripts/controller_group
mikepurvis: Using `ROS_NAMESPACE` could be nice because then you can do it once in your terminal and the ergonomics of the CLI usage are otherwise preserved (not needing an extra flag all the time). And you can get that pretty much for free I think by simply removing the leading slash from the param here.

Anyway, apart from that there isn't a single obvious way forward with this; since we don't have an immediate need, I'd just leave it.

controller_manager/scripts/controller_group
mikepurvis: I don't think it should be; `myargv` will have you covered— was more just acknowledging that that use case be considered.

controller_manager/scripts/controller_group
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env python
+# Provides quick access to the services exposed by controller_manager_msgs
+import rospy
+from argparse import ArgumentParser
+from controller_manager import controller_manager_interface
+from controller_manager_msgs.srv import ListControllers, ListControllersRequest
+
+
+def _read_group(group):
+    groups = rospy.get_param('controller_groups', {})
```
yli-cpr: Now using relative parameters, so that it respects ROS_NAMESPACE or roslaunch remapping. Confirmed it works.

controller_manager/CMakeLists.txt
```diff
@@ -46,7 +46,11 @@ install(DIRECTORY include/${PROJECT_NAME}/
 install(TARGETS ${PROJECT_NAME}
   LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION})
 
-install(PROGRAMS scripts/spawner scripts/unspawner scripts/controller_manager
+catkin_install_python(PROGRAMS
+  scripts/spawner
+  scripts/unspawner
+  scripts/controller_manager
+  scripts/controller_group
```
mikepurvis: Alpha ordering here.

controller_manager/scripts/controller_group
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env python
+# Provides quick access to the services exposed by controller_manager_msgs
```
mikepurvis: I know you copied this from `controller_manager`, but please use a full BSD header:

https://github.com/yli-cpr/ros_control/blob/f7c91ad13da18a3241990e3eeca2b34b092db45d/controller_manager/scripts/spawner#L1-L33

controller_manager/scripts/controller_group
mikepurvis: Consider adding a comment explaining this to future readers/devs.

controller_manager_tests/test/controller_manager_scripts.py
```diff
@@ -107,6 +112,33 @@ def test_scripts(self):
...
+        self.assertEqual(
+            run_cg('switch group2'),
+            'Stopped [\'my_controller1\'] successfully\nLoaded my_controller2\n'
+            'Started [\'my_controller2\'] successfully\n')
+
+        # switch to faulty group because my_controller1 conflicts with my_controller2
+        self.assertEqual(
+            run_cg('switch group3'),
+            'Stopped [\'my_controller3\'] successfully\n'
+            'Error when starting  [\'my_controller1\']\n')
```
mikepurvis: Shouldn't the switching happen atomically? Like, stop old and start new all in one go? That's one of the ideas of ros_control, is that even when you hot-switch controllers, it's possible for the system to never be uncontrolled.

So the output I'd expect here would be that the old controllers continue to run because the new config had a conflict and therefore wasn't loaded.

yli-cpr: I didn't use switch_controllers, and the reason is... probably silly. I am doing it with switch_controllers now.

controller_manager/scripts/controller_group
yli-cpr: sure

yli-cpr:
got it. fixing the error.

controller_manager_tests/test/controller_manager_scripts.test
yli-cpr: Now it uses switch_controllers service which will print this
```
[\ERROR] [\/dummy_app] [\1525462144.701348031]: \Could not switch controllers, due to resource conflict
```
The script will also print an error based on the false return value.

controller_manager/scripts/controller_group
```diff
@@ -0,0 +1,132 @@
...
+        if loaded_controllers.get(controller, False):
+            stop.append(controller)
+
+    # Load and start every desired controller if needed.
+    start = []
+    for controller in desired_controllers:
+        if controller not in loaded_controllers:
+            controller_manager_interface.load_controller(controller)
+        if not loaded_controllers.get(controller, False):
+            start.append(controller)
```
mikepurvis: Why is there a conditional here? If a controller that's needed couldn't be loaded, the whole operation should fail, right?

yli-cpr: I don't see switch_controller service loads the controller. So I have to load those controllers that haven't been loaded. Yeah. I should check `load_controller()`'s return value.

The second `if` is because switch_controllers service will fail if I try to start a controller that's already running, because we are using `Strict`

mikepurvis: Oh, got it— so you're figuring out the stops/starts required to get from the current state to the desired state, as opposed to just stop everything and start <desired>.

controller_manager_tests/test/controller_manager_scripts.test
mikepurvis: 👍

mikepurvis (approved):

mikepurvis:
This looks great to me. Ping @efernandez and @bmagyar for their thoughts.

controller_manager/scripts/controller_group
```diff
@@ -0,0 +1,135 @@
...
+    Return 2 sets of controllers. The first set includes controllers required for the given group,
+    and the second set includes the controllers that should be stopped if the given group is chosen.
+    '''
+    groups = rospy.get_param('controller_groups', {})
+    desired_controllers = set(groups.get(group, ()))
+    if not desired_controllers:
+        rospy.logerr('There is no controller defined in group {}'.format(group))
+        return None, None
+    all_controllers = set()
+    for controllers in groups.values():
```
efernandez: If groups is a `dict` you can also do `for controllers in groups.itervalues():`, which I think it's preferred over `values()`.

Then I'd update the comment below, b/c there won't be any `list`, to:
` # Add to a set to remove duplicates`

Or remove the comment from here and move it before `all_controllers = set()` and say `# Remove duplicates`.
I personally prefer this last option.

controller_manager/scripts/controller_group
```diff
@@ -0,0 +1,135 @@
...
+        # Convert list to set to remove duplicates.
+        all_controllers = all_controllers | set(controllers)
```
efernandez: I think `all_controllers.add(controllers)` is easier to read, at least for me.
I'm not familiar with the syntax you're using.

controller_manager/scripts/controller_group
```diff
@@ -0,0 +1,135 @@
...
+    return desired_controllers, all_controllers - desired_controllers
+
+
+def _get_loaded_controllers():
+    rospy.wait_for_service('controller_manager/list_controllers')
```
efernandez: Shouldn't we pass a `timeout=N`, where I think `N` could be 5s or so. Otherwise this could get block here and the user won't know what's going on.

If you add the timeout you need a try - except block catching `ROSException`.

controller_manager/scripts/controller_group
```diff
@@ -0,0 +1,135 @@
...
+    s = rospy.ServiceProxy('controller_manager/list_controllers', ListControllers)
+    resp = s.call(ListControllersRequest())
```
efernandez: This is the same as `resp = s(ListControllerRequest())` IIRC.

You also need a try - except block for this, in case the call fails for any of these reasons:
``` python
@raise TypeError: if request is not of the valid type (Message)
@raise ServiceException: if communication with remote service fails
@raise ROSInterruptException: if node shutdown (e.g. ctrl-C) interrupts service call
@raise ROSSerializationException: If unable to serialize
```

In this case, it's probably not a big deal, but I think `ServiceException` and `ROSInterruptException` should be caught. (or their base class, if any).

controller_manager/scripts/controller_group
```diff
@@ -0,0 +1,135 @@
...
+    for controller in resp.controller:
+        controllers[controller.name] = controller.state == 'running'
+        if controller.state not in ('running', 'stopped'):
+            rospy.logerr('Unknown controller state {}'.format(controller.state))
+    return controllers
+
+
+def _list_group(args):
+    groups = rospy.get_param('controller_groups', {})
+    print('Number of groups: {}'.format(len(groups)))
```
efernandez, line 71: Why do you use `print` here instead of `rospy.loginfo`?

In other parts of the code you use `rospy.logerr`, for instance.

controller_manager/scripts/controller_group
```diff
@@ -0,0 +1,135 @@
...
+    switch_parser = subparsers.add_parser('switch', help='switch to the specified group')
+    switch_parser.add_argument('group', help='name of the group to switch to')
+    switch_parser.set_defaults(func=_switch_group)
+
+    args = parser.parse_args(rospy.myargv()[1:])
+    args.func(args)
+
+
+if __name__ == '__main__':
+    _main()
```
efernandez, line 135: Uhm, I thought you had to `rospy.init_node('controller_group', sys.argv, anonymous=True)` to call services.
I think it's better to do so: the time will be initialized, and it'd allow to do remappings, although it's something we probably don't need here.

Another thing, that I think it's common, it's to have a try - except block to check for `KeyError`, so we detect when the user Ctrl+C the script.

controller_manager/src/controller_manager/controller_manager_interface.py
```diff
@@ -75,37 +75,28 @@ def unload_controller(name):
         return False
 
 def start_controller(name):
-    return start_stop_controllers([name], True)
+    return start_stop_controllers(start=[name])
```
efernandez: Or `start_controllers([name])`

controller_manager/src/controller_manager/controller_manager_interface.py
```diff
@@ -75,37 +75,28 @@ def unload_controller(name):
...
 
 def start_controllers(names):
-    return start_stop_controllers(names, True)
+    return start_stop_controllers(start=names)
 
 def stop_controller(name):
-    return start_stop_controllers([name], False)
+    return start_stop_controllers(stop=[name])
```
efernandez: Or `stop_controllers([name])`

controller_manager/src/controller_manager/controller_manager_interface.py
```diff
@@ -75,37 +75,28 @@ def unload_controller(name):
...
-    else:
-        stop = names
     resp = s.call(SwitchControllerRequest(start, stop, strictness))
     if resp.ok == 1:
-        if st:
-            print "Started {} successfully".format(names)
-        else:
-            print "Stopped {} successfully".format(names)
+        if start:
+            print "Started {} successfully".format(start)
```
efernandez: Not really from this PR, but you could add parenthesis to `print`. Here and below

controller_manager_tests/test/controller_manager_scripts.py
```diff
@@ -107,6 +112,32 @@ def test_scripts(self):
 
         self.assertEqual(run_cm('reload-libraries'), reload_response)
 
+        # controller_group
+        # list
+        self.assertEqual(
+            run_cg('list'),
+            'Number of groups: 3\n  group1\n      my_controller1\n      my_controller3\n'
```
efernandez: Could you also move group1 to its own line, please? It'd be easier to read.

efernandez (changes requested):
LGTM, only minor comments.

controller_manager/src/controller_manager/controller_manager_interface.py
bmagyar: Make sure to put this in a separate commit please, I'd cherry pick it to kinetic

controller_manager/scripts/controller_group
bmagyar: Or let it err out with the exception, although less pretty, that's not bad either.

controller_manager/scripts/controller_group
yli-cpr: The problem is itervalues() is not supported by Python 3. Also the performance gain is trivial for this use case. Will move the comment.

controller_manager/scripts/controller_group
yli-cpr: `controllers` is a list of objects. I can change it to `all_controllers.union(set(controllers))` or add another loop.

controller_manager/scripts/controller_group
yli-cpr: Was following controller_manager_interface. I will add the a timeout and let it crash with the exception.

controller_manager/scripts/controller_group
yli-cpr: When they are not caught, the exception will be logged. This script isn't supposed to keep running. What value does catching them give us?

controller_manager/scripts/controller_group
yli-cpr: Because in this case it is supposed to print the list to stdout, like printing usages. It shouldn't go through any logger IMHO.

controller_manager/scripts/controller_group
yli-cpr: I did that for a script that uses Twisted to stop reactor. I think it should be done when necessary. If I add it for this script, I don't know what to do with with exception except printing something. Would leave it to python default behavior.

controller_manager/src/controller_manager/controller_manager_interface.py
yli-cpr: Was trying to follow the style of `controller_manager_interface.py`. Also `_controllers` doesn't give more information because the function is called `start_stop_controllers()`

controller_manager/src/controller_manager/controller_manager_interface.py
yli-cpr: @bmagyar Do you mean `print(msg)` style shouldn't be in kinetic?

controller_manager_tests/test/controller_manager_scripts.py
yli-cpr: Sure

controller_manager/scripts/controller_group
efernandez: I think I prefer `all_controllers.union(set(controllers))`, but I leave it up to you.

controller_manager/scripts/controller_group
efernandez: Ok, nvm.

Sounds good.

controller_manager/scripts/controller_group
efernandez: I see, makes sense.

controller_manager/scripts/controller_group
efernandez: Ok. As long as the user can Ctrl+C to stop the script if needed. It looks good to me.

controller_manager/src/controller_manager/controller_manager_interface.py
efernandez: The reason I'd call `start_controllers([name])` is b/c if in the future sth is added there, it's better to have a single entry point. I think we'll always handle a single element the same way as a list of a single element.

controller_manager/scripts/controller_group
efernandez: I think you're right, nvm.

controller_manager/src/controller_manager/controller_manager_interface.py
efernandez: I think he wants to cherry-pick only that style/deprecation change into `kinetic-devel`.
This PR is targeting `modelodic-devel`, btw.

controller_manager/scripts/controller_group
yli-cpr: will do `union`

controller_manager/src/controller_manager/controller_manager_interface.py
yli-cpr: Got it. Sure.

controller_manager/src/controller_manager/controller_manager_interface.py
yli-cpr: I see.

controller_manager/scripts/controller_group
yli-cpr: actually there is an `update` method for `set` which accepts any iterable as input. will use that.

bmagyar (changes requested):
Great work! I have one request before putting a green light on this: Please add documentation here: http://wiki.ros.org/controller_manager

yli-cpr:
@bmagyar The wiki page has been updated.
Ping @efernandez for another look.

bmagyar (approved):

bmagyar:
@efernandez feel free to merge if you greenlight it too!

efernandez (approved):

mikepurvis:
Awesome, thanks for your work on this, @yli-cpr! A possible follow-on at some point might be to add a `controller_group status` verb or similar, which would examine the currently running controllers and tell you which (if any) group is currently active.

yli-cpr:
That's good idea. Will add it to todo list

PR #342, opened 2018-07-04
Added quotes for controller name and controller type in warnings and errors

jgueldenstein:
This PR provides some consistency in the warnings and errors printed by the controller manager.
Having quotes around the controller requested to be spawned is especially useful when the controller name was left empty or contains whitespace.

It would also be nice to have this in kinetic too

@v4hn

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -207,7 +207,7 @@ bool ControllerManager::loadController(const std::string& name)
     }
     catch (const std::runtime_error &ex)
     {
-      ROS_ERROR("Could not load class %s: %s", type.c_str(), ex.what());
+      ROS_ERROR("Could not load class '%s': %s", type.c_str(), ex.what());
     }
   }
   else
@@ -234,11 +234,11 @@ bool ControllerManager::loadController(const std::string& name)
     initialized = c->initRequest(robot_hw_, root_nh_, c_nh, claimed_resources);
   }
   catch(std::exception &e){
-    ROS_ERROR("Exception thrown while initializing controller %s.\n%s", name.c_str(), e.what());
+    ROS_ERROR("Exception thrown while initializing controller '%s'.\n%s", name.c_str(), e.what());
     initialized = false;
   }
   catch(...){
-    ROS_ERROR("Exception thrown while initializing controller %s", name.c_str());
+    ROS_ERROR("Exception thrown while initializing controller '%s'", name.c_str());
     initialized = false;
   }
   if (!initialized)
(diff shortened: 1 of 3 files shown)


bmagyar:
LGTM :+1: 

@ipa-mdl please review and merge if ok for you too.

bmagyar (approved):

bmagyar (changes requested):
Actually... I think the same should be applied to the Python interfaces as well.

mathias-luedtke:
The migration to `rosconsole` breaks the corresponding [test](https://github.com/ros-controls/ros_control/blob/melodic-devel/controller_manager_tests/test/controller_manager_scripts.py).
Does this even work properly? The script does not start a node ..

v4hn:
Why does changing to `rospy.loginfo` result in
`controller_manager_tests/EffortTestController' missing?
Or do I misunderstand the CI output?

The code around the `print`'s in question calls ROS services.
That might even be possible without being a node, but
I would still expect that users only use this within a ROS node.

What do you propose to resolve this?

bmagyar:
Could we undo the rosconsole migration in the python parts please? 
This library supports command line tools and as such it creates it's output directly on stdout. 

@v4hn the tests are failing because we specifically test for the right text output.

v4hn:
Makes sense.

mathias-luedtke:
> Could we undo the rosconsole migration in the python parts please?

As an alternative the tests could be adapted to strip the rosconsole prefix (with the varying time stamp..).

jgueldenstein:
I will undo the rosconsole part and adapt the tests as soon as I get to it.

jgueldenstein:
I had some time to look at the test results. I do not know why the occurring error
```
[controller_manager_tests.rosunit-controller_manager_interface_test/test_scripts][passed]
[controller_manager_tests.rosunit-controller_manager_scripts_test/test_scripts][FAILURE]
'controller_manager_tests/EffortTestController' not found in []
  File "/usr/lib/python2.7/unittest/case.py", line 329, in run
    testMethod()
  File "/root/catkin_ws/src/ros_control/controller_manager_tests/test/controller_manager_scripts.py", line 56, in test_scripts
    self.assertIn('controller_manager_tests/EffortTestController', listed_types)
  File "/usr/lib/python2.7/unittest/case.py", line 804, in assertIn
    self.fail(self._formatMessage(msg, standardMsg))
  File "/usr/lib/python2.7/unittest/case.py", line 410, in fail
    raise self.failureException(msg)
```
starting in line 4169 could be related to these changes in the code.
Please have a look.

jgueldenstein:
ping

bmagyar:
You haven't undone the rosconsole parts, this means that the tests that are capturing `stdout` will get empty strings from the subprocess calls. 

I've also spotted a few other places where I had to add `'`-s and adjusted the expected strings in the test.

v4hn:
Thanks @bmagyar !
Tests pass now.

bmagyar (approved):

bmagyar:
@ipa-mdl could you please give a wee look and merge if all ok?

jgueldenstein:
Ok thank you for clarifying, I was just confused that a test which seemed totally unrelated failed.
Also thank you for fixing.

bmagyar:
Extending to: @ipa-mdl or @graiola

PR #345, opened 2018-07-10
added C++11 flag

geoffviola:
I needed this change for lunar on Ubuntu 16.04

### DIFF
--- a/combined_robot_hw/CMakeLists.txt
+++ b/combined_robot_hw/CMakeLists.txt
@@ -1,6 +1,8 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(combined_robot_hw)
 
+set(CMAKE_CXX_STANDARD 11)
+
 find_package(catkin REQUIRED COMPONENTS
   hardware_interface
   pluginlib
--- a/combined_robot_hw_tests/CMakeLists.txt
+++ b/combined_robot_hw_tests/CMakeLists.txt
@@ -1,6 +1,8 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(combined_robot_hw_tests)
 
+set(CMAKE_CXX_STANDARD 11)
+
 find_package(catkin REQUIRED COMPONENTS
   combined_robot_hw
   controller_manager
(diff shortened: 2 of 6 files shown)


bmagyar:
Wouldn't this remove functionality on platforms where gcc has C++14 support enabled by default? 
You could check if `CXX_STANDARD == 98` and only then set it `CMAKE_CXX_STANDARD 11`
https://cmake.org/cmake/help/v3.1/prop_tgt/CXX_STANDARD.html#prop_tgt:CXX_STANDARD

@ipa-mdl what do you think? Something like this could indeed help supporting compile-from-source on Ubuntu < 18.04.

geoffviola:
I don't think it will remove the functionality. According to CMake "This property specifies the C++ standard whose features are requested to build this target". It's probably a good idea to test it anyway.

mathias-luedtke:
I don't see the need to add this flag. We branched it out for melodic. `kinetic-devel` works perfectly fine on `lunar`.
If we really want to set this flag, I would set it to c++ 14 (ROS melodic default).

Please note, that setting flag leads to ABI breaks if mixed C++03 code.

geoffviola:
Are you using Ubuntu 18.04? What version of GCC? C++14 was the default in GCC 6.1.

mathias-luedtke:
> Please note, that setting the c++11 flag leads to ABI breaks with C++03 code.

I don't want to support the `melodic-devel` branch on `xenial` (or `trusty`).
(In accordance with [REP 3](http://www.ros.org/reps/rep-0003.html))

mathias-luedtke:
@geoffviola: Why can't you switch to the `kinetic-deve` branch?
@bmagyar l think we should point out in the README that the master branch is meant for melodic only and that the `kinetic-devel` branch supports  `kinetic` and `lunar`.

geoffviola:
Sorry, I thought it was not going to be backwards compatible. The kinetic branch seems like it would work better for me.

bmagyar:
Thanks @ipa-mdl for pointing to `REP 3`, it makes it clear. 
@geoffviola , thanks for bringing up this point though, we'll update the readme with a note for Lunar.

PR #357, opened 2018-09-17
Add method to check for error in robotHW component

bochen87:
I'm proposing to add a method to check for errors in the hardware components, especially when having multiple HW components, it becomes quite cumbersome otherwise.

### DIFF
--- a/combined_robot_hw/include/combined_robot_hw/combined_robot_hw.h
+++ b/combined_robot_hw/include/combined_robot_hw/combined_robot_hw.h
@@ -100,6 +100,11 @@ public:
    */
   virtual void write(const ros::Time& time, const ros::Duration& period);
 
+  /**
+   * Checks whether the robot HW components have errors
+   */
+  virtual bool hasError();
+
 protected:
   ros::NodeHandle root_nh_;
   ros::NodeHandle robot_hw_nh_;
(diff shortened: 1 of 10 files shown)


bmagyar:
Could you please demonstrate an example use-case, perhaps through some unit tests?

bochen87:
@bmagyar added a very basic test to it. I guess could also be very useful in conjunction with https://github.com/ros-controls/ros_control/pull/294 to then automatically attempt to recover / stop

mathias-luedtke:
Who would call this `hasError` function and why?

bochen87:
At the moment I use this in the main loop and then attempt to recover / stop if it returns true.

mathias-luedtke:
> attempt to recover / stop if it returns true.

This is the interesting part.. I believe there are no interfaces for this yet.
In `ros_canopen` I poll and command the hardware directly.

bochen87:
No, there is no interface for that yet - that's why I would like to see https://github.com/ros-controls/ros_control/pull/294 being extended to combined_robot_hw and also merged.
Maybe add the extension of stop / recover to combined_robot_hw and the error checking interface in one PR?

bochen87:
@bmagyar @ipa-mdl any thoughts?

jlack1987:
Would just like to throw this out there in case nobody has considered it, but it appears the API is converging on an [orocos component](http://www.orocos.org/stable/documentation/rtt/v2.x/doc-xml/orocos-components-manual.html) like design

bochen87:
> @bmagyar @ipa-mdl any thoughts?

Bump

PR #364, opened 2018-10-18
Spawner check for currently running controllers on shutdown

megantuttle:
This PR improves controller shutdown by getting the list of currently running controllers before attempting to stop and unload them.  Previously, shutdown simply stopped and unloaded the controllers that were spawned on launch. This behavior caused errors if a different controller was running at the time shutdown was requested.

### DIFF
--- a/controller_manager/scripts/spawner
+++ b/controller_manager/scripts/spawner
@@ -41,6 +41,7 @@ import yaml
 from controller_manager_msgs.srv import *
 from std_msgs.msg import *
 import argparse
+import controller_manager.controller_manager_interface
 
 loaded = []
 
@@ -49,27 +50,37 @@ load_controller_service = ""
 switch_controller_service = ""
 unload_controller_service = ""
 
+counter = 0
+
 def shutdown():
     global loaded,unload_controller_service,load_controller_service,switch_controller_service
 
     rospy.loginfo("Shutting down spawner. Stopping and unloading controllers...")
 
     try:
-        # unloader
-        unload_controller = rospy.ServiceProxy(unload_controller_service, UnloadController)
+        # get list of running controllers
+        _running_controllers = []
+        _loaded_controllers = controller_manager.controller_manager_interface.list_controllers()
+        for c in _loaded_controllers:
+            if c.state == "running":
+                _running_controllers.append(c.name)
 
-        # switcher
+        # create unloader and switcher
+        unload_controller = rospy.ServiceProxy(unload_controller_service, UnloadController)
         switch_controller = rospy.ServiceProxy(switch_controller_service, SwitchController)
 
         rospy.loginfo("Stopping all controllers...");
-        switch_controller([], loaded, SwitchControllerRequest.STRICT)
+        switch_controller([], _running_controllers, SwitchControllerRequest.STRICT)
         rospy.loginfo("Unloading all loaded controllers...");
-        for name in reversed(loaded):
+        for name in reversed(_running_controllers):
             rospy.logout("Trying to unload %s" % name)
             unload_controller(name)
             rospy.logout("Succeeded in unloading %s" % name)
     except (rospy.ServiceException, rospy.exceptions.ROSException) as exc:
         rospy.logwarn("Controller Spawner error while taking down controllers: %s"  % (exc))
+        counter += 1
+        if counter < 3:
+            shutdown()
 
 def parse_args(args=None):
     parser = argparse.ArgumentParser(description='Controller spawner')
@@ -190,9 +201,6 @@ def main():
             rospy.logerr("Failed to load %s" % name)
 
     rospy.loginfo("Controller Spawner: Loaded controllers: %s" % ', '.join(loaded))
-
-    if rospy.is_shutdown():
-        return
 
     # start controllers is requested
     if autostart:
--- a/controller_manager/scripts/spawner
+++ b/controller_manager/scripts/spawner
@@ -63,14 +63,17 @@ def shutdown():
             if c.state == "running":
                 _running_controllers.append(c.name)
 
+        # determine if the loaded controllers are still running
+        _running_loaded_controllers = list(set(_running_controllers) & set(loaded))
+
         # create unloader and switcher
         unload_controller = rospy.ServiceProxy(unload_controller_service, UnloadController)
         switch_controller = rospy.ServiceProxy(switch_controller_service, SwitchController)
 
-        rospy.loginfo("Stopping all controllers...");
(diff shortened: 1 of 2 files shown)


controller_manager/src/controller_manager/controller_manager_interface.py
```diff
@@ -108,13 +102,13 @@ def start_stop_controllers(names, st):
     resp = s.call(SwitchControllerRequest(start, stop, strictness))
     if resp.ok == 1:
         if st:
-            print("Started {} successfully".format(names))
```
bmagyar: I think your PR needs rebasing. Please don't change the prints.

bmagyar (changes requested):
I see the general idea with the PR but it is a little off. The spawner should only be responsible for stopping/unloading the controllers it loaded/started.

You should take the intersection of currently running vs loaded/started controller and only act on those.

controller_manager/src/controller_manager/controller_manager_interface.py
megantuttle: Why should the spawner not be responsible for shutting down all running controllers?  I don't see a problem with making that change, I'm just trying to understand it.

And I'm not sure how the prints ended up being changed, it must have been my text editor.  I'll fix those.

mathias-luedtke (changes requested):
The `spawner` should not deal with any controller that it was not configured for.

> Previously, shutdown simply stopped and unloaded the controllers that were spawned on launch. This behavior caused errors if a different controller was running at the time shutdown was requested.

If this really causes errors, there is a bug that needs to be fixed.
The current PR would start controllers that have been stopped (or even unloaded) in between.

megantuttle:
Error output from the original issue I was trying to resolve:
`[ros_control_controller_spawner-2] killing on exit`
`[INFO] [1539901011.740469]: Shutting down spawner. Stopping and unloading controllers...`
`[INFO] [1539901011.741018]: Stopping all controllers...`
`[ERROR] [1539901011.743983745]: Could not stop controller 'pos_based_pos_traj_controller' since it is not running`
`[INFO] [1539901011.744493]: Unloading all loaded controllers...`
`[INFO] [1539901011.744950]: Trying to unload pos_based_pos_traj_controller`
`[INFO] [1539901011.782248]: Succeeded in unloading pos_based_pos_traj_controller`
`[INFO] [1539901011.782680]: Trying to unload force_torque_sensor_controller`
`[ERROR] [1539901011.785208211]: Could not unload controller with name force_torque_sensor_controller because it is still running`
`[INFO] [1539901011.785470]: Succeeded in unloading force_torque_sensor_controller`
`[INFO] [1539901011.785781]: Trying to unload joint_state_controller`
`[ERROR] [1539901011.788846453]: Could not unload controller with name joint_state_controller because it is still running`
`[INFO] [1539901011.789072]: Succeeded in unloading joint_state_controller`

I am running Kinetic on Ubuntu 16.04, and using the `ur_modern_driver` package to control a UR3.  I switch from the `pos_based_pos_traj_controller` (started by default) to the `vel_based_pos_traj_controller` using the `switch_controllers` service shortly after starting up the system by calling `ur3_ros_control.launch` and `ur3_moveit_planning_execution.launch`.

Why is `spawner` only intended to shut down the controllers it started?  Why shouldn't it stop any controllers that are running when shutdown is requested?

controller_manager/src/controller_manager/controller_manager_interface.py
```diff
@@ -1,18 +1,15 @@
 #! /usr/bin/env python
```
mathias-luedtke: please do not change this file

controller_manager/scripts/spawner
```diff
@@ -41,6 +41,7 @@ import yaml
 from controller_manager_msgs.srv import *
 from std_msgs.msg import *
 import argparse
+import controller_manager.controller_manager_interface
```
mathias-luedtke: please  use the same approach for listing controllers as used for the other services.

controller_manager/scripts/spawner
```diff
@@ -191,9 +200,6 @@ def main():
 
     rospy.loginfo("Controller Spawner: Loaded controllers: %s" % ', '.join(loaded))
 
-    if rospy.is_shutdown():
```
mathias-luedtke, line 194: whiy did you remove this block?

mathias-luedtke (changes requested):

mathias-luedtke:
> Why is spawner only intended to shut down the controllers it started? Why shouldn't it stop any controllers that are running when shutdown is requested?

The intended use case for `spawner` is to provide node that manages a (sub)set of controllers.
So the life cycle of the controllers is bound to the lifecycle of the node.
For example, a user could prepare different launch file for different control set-ups.
`spawner` "owns" the controllers, so these should better not stopped manually.

Instead of adding the list call, it might be easier to just use `BEST_EFFORT` for the stopping call.

controller_manager/scripts/spawner
mathias-luedtke: (or do not call the service at all, see https://github.com/ros-controls/ros_control/pull/364#issuecomment-431430987)

controller_manager/scripts/spawner
megantuttle: It seemed redundant since a shutdown hook is defined, and `rospy.spin()` simply keeps python from exiting until `rospy.core.is_shutdown()` is set (see [rospy/client.py](https://github.com/ros/ros_comm/blob/kinetic-devel/clients/rospy/src/rospy/client.py#L118)).

controller_manager/src/controller_manager/controller_manager_interface.py
megantuttle: Reverted in ee4cf92

controller_manager/scripts/spawner
megantuttle: Done. See ee4cf92

bmagyar:
Just to give some context, on a humanoid we'd usually have a separate set
of controllers for arms, legs/mobile base managed by different spawners but
within the same controller_manager as they belong to the same hw comm
system.

bmagyar:
@megantuttle will you have time to revamps this or would you prefer to close this PR? 
IMHO this PR still has the potential to make shutdown more robust to changes (if the spawned controller was stopped elsewhere).

megantuttle:
> @megantuttle will you have time to revamps this or would you prefer to close this PR?
> IMHO this PR still has the potential to make shutdown more robust to changes (if the spawned controller was stopped elsewhere).

@bmagyar I believe I've addressed all of the comments here, am I missing something?

PR #374, opened 2019-02-01
remove unused pthread.h

kejxu:
```
#include <pthread.h>
```
would cause build break on Windows since the header is not available. since it's not used in the code, remove


mathias-luedtke (approved):

bmagyar:
Good catch, thanks!

PR #377, opened 2019-03-01
fix install destination

kejxu:
and add comment
http://docs.ros.org/melodic/api/catkin/html/howto/format2/building_libraries.html
http://docs.ros.org/melodic/api/catkin/html/howto/format2/building_executables.html


kejxu:
@ipa-mdl, these changes in `ros_control` were left out from the last patch, could we get a quick merge for this one? =)

btw, sorry for the spam!

mathias-luedtke (approved):

PR #382, opened 2019-05-09
Fix exception module

garaemon:
ROSServiceIOException is under rosservice module.
http://docs.ros.org/lunar/api/rosservice/html/rosservice.ROSServiceIOException-class.html


bmagyar (changes requested):
Is this causing a failure when you are using it?

Could you please try adding a unit test that reproduces a failure?

bmagyar (approved):

PR #384, opened 2019-07-19
Update testing repository URL

matthew-reynolds:
In response to the security breach last month, all packages should move from using the `ros-shadow-fixed` repository to the new `ros-testing` repository.

See [@tfoote's post on the discourse](https://discourse.ros.org/t/new-gpg-keys-deployed-for-packages-ros-org/9454).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ env:
     - ROSINSTALL_FILENAME=ros_control.rosinstall
     - ROS_PARALLEL_TEST_JOBS=-j1
   matrix:
-    - ROS_DISTRO=melodic ROS_REPO=ros-shadow-fixed
+    - ROS_DISTRO=melodic ROS_REPO=ros-testing
 
 install:
   - git clone https://github.com/ros-industrial/industrial_ci.git .ci_config


mathias-luedtke:
Thanks for your effort, but this is not really needed..
However, it might avoid  some confusion.

> all packages should move from using the ros-shadow-fixed repository to the new ros-testing repository.

That's why we aliased it in `industrial_ci` (see my [post](https://discourse.ros.org/t/industrial-ci-migration-needed-for-recent-and-future-changes/9462)).

matthew-reynolds:
> However, it might avoid some confusion.

That was my main goal here. Even though `industrial_ci` already redirects to the correct repo, we still ought to move to `ros-testing` to remove references to the old repo to avoid confusion, especially among new users who might look at popular packages for examples on how to implement CI.

mathias-luedtke:
@bmagyar: Is there any reason why we don't run tests for `ROS_REPO=ros` as well?

bmagyar:
The bulk of the framework lives in this repo, we could run a test pipeline against `ros` too, but it wouldn't gain us much, only add a lot of extra time. The aliasing in industrial_ci is great, however just as with typedef-ing boost types, we eventually want to remove refs to the unused things, not only map. I'm happy to go ahead with renaming

bmagyar (approved):

PR #329, opened 2018-05-01
Single controller template (ROS Melodic)

mikepurvis:
Successor to #302. Not a clean rebase for Melodic, so just making a separate PR for comparison purposes.

See discussion in the original PR. IMO this is a worthwhile change to get in for Melodic— it deprecates `MultiInterfaceController`, but otherwise should be fully backward compatible with source.

PR #387, opened 2019-08-01
Variadic Controller class + Variadic CompositeController class

mikepurvis:
This is a combination of changes previously proposed in #301, #302, and #329. The two additions, both made possible by ROS Kinetic+ being on C++11, are:

- Instead of `Controller` taking a single Interface template parameter, it now takes an arbitrary number of them. This should be backward source compatible with previous `Controller` use, and because it's all headers, there's no ABI breakage. This also means that `MultiInterfaceController` no longer needs to exist, and can be deprecated at a future point (for now it's been made a pass-thru to `Controller`).
- A new `CompositeController` template has been added, which permits combining controllers, and in particular modifying behaviour when the base controller(s) supply extension points in the form of virtual functions.

Clearpath have been using both of these features for 18+ months and found them useful.

### DIFF
--- a/controller_interface/include/controller_interface/composite_controller.h
+++ b/controller_interface/include/controller_interface/composite_controller.h
@@ -38,6 +38,56 @@
 namespace controller_interface
 {
 
+/**
+ * \brief Merge together two ClaimedResources vectors, detecting whether the
+ * any of the merged-in resources were already present. The merge will be
+ * fully executed even if the function returns false (eg, indicating a conflict).
+ *
+ * \param[out] target ClaimedResources to merge resources into.
+ *
+ * \param source ClaimedResources to draw resources from.
+ *
+ * \return true if no resources were duplicated between the target and source.
+ */
+static bool mergeClaimedResources(ControllerBase::ClaimedResources& target,
+                                  const ControllerBase::ClaimedResources& source)
+{
+  bool conflict = false;
+  for (const auto& claimed : source)
+  {
+    // Check if there's a ClaimedResource of this interface type already in the target.
+    bool resource_in_target = false;
+    for (auto& target_claimed : target)
+    {
+      if (target_claimed.hardware_interface == claimed.hardware_interface)
+      {
+        resource_in_target = true;
+
+        size_t size_before = target_claimed.resources.size();
+        target_claimed.resources.insert(claimed.resources.begin(),
+                                        claimed.resources.end());
+
+        // Since it's a set, if the resulting size isn't big enough, we know that
+        // one or more of the inserted items was already present.
+        if (target_claimed.resources.size() < size_before + claimed.resources.size())
+        {
+          conflict = true;
+        }
+
+        break;
+      }
+    }
+
+    if (!resource_in_target)
+    {
+      // There isn't, so append the whole ClaimedResource to the vector.
+      target.push_back(claimed);
+    }
+  }
+  return !conflict;
+}
+
+
 template <typename... Controllers>
 class CompositeController: public virtual ControllerBase, public Controllers...
 {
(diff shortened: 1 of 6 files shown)


controller_interface/include/controller_interface/composite_controller.h
```diff
@@ -0,0 +1,285 @@
...
+  {
+    return callInitRequest<Controller1>(robot_hw, root_nh, controller_nh, all_claimed_resources) &&
+      callInitRequest<Controller2, More...>(robot_hw, root_nh, controller_nh, all_claimed_resources);
+  }
+
+  CompositeController(const CompositeController& c);
+  CompositeController& operator =(const CompositeController& c);
+};
+
+} // namespace
```
mathias-luedtke, line 283: please add the name of the namespace as well

controller_interface/include/controller_interface/controller.h
```diff
@@ -85,65 +209,106 @@ class Controller: public virtual ControllerBase
...
+  Controller& operator =(const Controller& c);
 
+  // Everything other than this line actually compiles just fine under gcc without -std=c++11,
+  // barring warnings about varidic templates being unsupported. This guard can be removed in Melodic
+  // when the compiler is c++14 by default.
+  static_assert(sizeof...(Interfaces) >= 1, "Controller must have at least one hardware interface.");
 };
 
-}
+} // namespace
```
mathias-luedtke, line 312: please add the name of the namespace as well

controller_interface/include/controller_interface/internal/robothw_interfaces.h
```diff
@@ -63,7 +63,7 @@ inline std::string enumerateElements(const T&           val,
 }
 
 
-template <typename T>
+template <class T>
```
mathias-luedtke, line 66: I guess, you changed this for consistency reasons?

mathias-luedtke (changes requested):
Useful indeed!

I have no doubt that the code works, but a test for `CompositeController` should be added in `controller_manager_tests` to maintain it properly.

controller_interface/include/controller_interface/internal/robothw_interfaces.h
mikepurvis: To be honest, I don't remember the logic for this change, but I don't think it matters; I'm happy for us to consistently use `typename` or `class`, and can revert this part of the change if the decision is for `typename`. I think the only time it matters, `typename` is preferred (eg in ab14ea48d73a1037f79686cf64f3379f9eac0a92).

mikepurvis:
@ipa-mdl Thanks for looking. I agree on some kind of test; I think in the past when working on this, I got hung up on what exactly such a test should do. Given that the resource merging is already covered by a unit test, is it enough to just demonstrate something which compiles and maybe passes a trivial smoke test? Or should there be a full ControllerA and ControllerB, and then CompositeControllerAB, with a complete lifecycle and verification of the init/start/stop methods being properly called?

I'm totally down for that approach, I think my concern is just about keeping the test short and simple enough that it can be easily grokked.

mathias-luedtke:
>  Given that the resource merging is already covered by a unit test, is it enough to just demonstrate something which compiles and maybe passes a trivial smoke test?

IMHO something like https://github.com/ros-controls/ros_control/commit/90f89e05eb87da69c9a48c31690b4ac954a23fce should do for now.

controller_interface/include/controller_interface/internal/robothw_interfaces.h
mathias-luedtke: > and can revert this part of the change if the decision is for typename

The other templates use `class` in their definitions as well, so it is fine for me.
`class` seems to be mandatory for nested templates (C++14 and older).

bmagyar: No strong opinion on this, do it at your leasure.

PR #391, opened 2019-10-21
Multi switch

jordan-palacios:
See the original discussion here: https://github.com/pal-robotics-forks/ros_control2/pull/5


bmagyar:
@toliver perhaps someone from Shadow may want to chime in?

bmagyar (approved):
I removed an unrelated commit that shouldn't be on this PR (hence the force-push) but other than that this PR looks good for me.

I'd like to get some user-level documentation for this though, some simple example code which is explained but I think we can do that as follow-up work.

controller_interface/test/controller_base_test.cpp
```diff
@@ -0,0 +1,313 @@
...
+// Copyright (C) 2019
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//   * Redistributions of source code must retain the above copyright notice,
+//     this list of conditions and the following disclaimer.
+//   * Redistributions in binary form must reproduce the above copyright
+//     notice, this list of conditions and the following disclaimer in the
+//     documentation and/or other materials provided with the distribution.
+//   * Neither the name of Pal Robotics S.L. nor the names of its
```
toliver, line 11: If Pal Robotics owns the Copyright maybe it should be clear in line 2?

controller_interface/test/controller_base_test.cpp
```diff
@@ -0,0 +1,313 @@
+///////////////////////////////////////////////////////////////////////////////
+// Copyright (C) 2019
```
bmagyar:
```suggestion
// Copyright (C) 2019, Pal Robotics S.L.
```

controller_interface/test/controller_base_test.cpp
bmagyar: good point! I'll add some suggestions

controller_manager/test/hwi_update_test.cpp
```diff
@@ -0,0 +1,520 @@
+///////////////////////////////////////////////////////////////////////////////
+// Copyright (C) 2019
```
bmagyar:
```suggestion
// Copyright (C) 2019, Pal Robotics S.L.
```

bmagyar (changes requested):

bmagyar (approved):

PR #392, opened 2019-11-19
Flexible actuator and joint interfaces

bmagyar:
This PR is an attempt to extend the available interfaces beyond position, velocity and effort, in a dynamic manner. Unfortunately this trinity of interfaces is heavily hardcoded troughout the framework, making this work of a very significant size.

* [x] Make ActuatorData templated on interface types
* [x] Make ActuatorStateHandle templated on interface types
* [x] TransmissionHandle will become templated on the components of ActuatorData, instead of holding a fixed ActuatorData actuator_data_; field. Add new tests.
* [ ] Adapt transmission parser to handle non-predefined interfaces


olivier-stasse:
@bmagyar How can we help on this one ?

matthew-reynolds:
Since this is planned for `ros2_control`, and Noetic is coming up quickly, is this still a goal for `ros_control`?

bmagyar:
We could if there's a need for it but definitely not the approach I took on this PR, it is now heavily conflicted and can expect only more conflicts with such a huge bulk (4000 lines of changes).
I'll close it for now and we can reference it later.

PR #395, opened 2019-12-02
Extend interfaces melodic

bmagyar:
- Adds `absolute_position` and `torque_sensor` data types to `JointStateHandle` and `ActuatorStateHandle`, as well as the corresponding transformations to `Transmission`s
 - Changes the way velocity joint saturation limits are calculated: It now uses previous commanded velocity rather than measured velocity.
 - Exposes raw data pointers in `JointStateHandle` and `JointCommandHandle`. Previously only available in actuator handles.
 - Adds a new `JointModeHandle` for switching a hardware interface between different controller modes

### DIFF
--- a/hardware_interface/include/hardware_interface/actuator_state_interface.h
+++ b/hardware_interface/include/hardware_interface/actuator_state_interface.h
@@ -37,7 +37,11 @@
 namespace hardware_interface
 {
 
-/** A handle used to read the state of a single actuator. */
+/** 
+ * \brief A handle used to read the state of a single actuator.
+ * Currently, position, velocity and effort fields are required 
+ * while absolute position and torque sensors are optional.
+ */
 class ActuatorStateHandle
 {
 public:
@@ -49,7 +53,6 @@ public:
    * \param vel A pointer to the storage for this actuator's velocity
    * \param eff A pointer to the storage for this actuator's effort (force or torque)
    */
-
   ActuatorStateHandle(const std::string& name, const double* pos, const double* vel, const double* eff)
     : name_(name), pos_(pos), vel_(vel), eff_(eff), absolute_pos_(0), torque_sensor_(0)
   {
@@ -65,9 +68,50 @@ public:
     {
       throw HardwareInterfaceException("Cannot create handle '" + name + "'. Effort data pointer is null.");
     }
-
   }
 
+  /**
+   * \param name The name of the actuator
+   * \param pos A pointer to the storage for this actuator's position
+   * \param vel A pointer to the storage for this actuator's velocity
+   * \param eff A pointer to the storage for this actuator's effort (force or torque)
+   * \param absolute_pos A pointer to the storage for this actuator's absolute position encoder
+   * \param torque_sensor A pointer to the storage for this actuator's torque sensor
+   */
+  ActuatorStateHandle(const std::string& name, const double* pos, const double* vel, const double* eff,
+                      const double* absolute_pos, const double* torque_sensor)
+    : name_(name), pos_(pos), vel_(vel), eff_(eff), absolute_pos_(absolute_pos), torque_sensor_(torque_sensor)
+  {
+    if (!pos)
+    {
+      throw HardwareInterfaceException("Cannot create handle '" + name + "'. Position data pointer is null.");
+    }
+    if (!vel)
+    {
+      throw HardwareInterfaceException("Cannot create handle '" + name + "'. Velocity data pointer is null.");
+    }
+    if (!eff)
+    {
+      throw HardwareInterfaceException("Cannot create handle '" + name + "'. Effort data pointer is null.");
+    }
+
+    if (!absolute_pos)
+    {
+      throw HardwareInterfaceException("Cannot create handle '" + name + "'. Absolute data pointer is null.");
+    }
+    if (!torque_sensor)
+    {
+      throw HardwareInterfaceException("Cannot create handle '" + name + "'. Torque sensor data pointer is null.");
+    }
+  }
+
+  /**
+   * \param name The name of the actuator
+   * \param pos A pointer to the storage for this actuator's position
+   * \param vel A pointer to the storage for this actuator's velocity
+   * \param eff A pointer to the storage for this actuator's effort (force or torque)
+   * \param absolute_pos A pointer to the storage for this actuator's absolute position encoder
+   */
   ActuatorStateHandle(const std::string& name, const double* pos, const double* vel, const double* eff,
                       const double* absolute_pos)
     : name_(name), pos_(pos), vel_(vel), eff_(eff), absolute_pos_(absolute_pos), torque_sensor_(0)
@@ -88,12 +132,18 @@ public:
     {
       throw HardwareInterfaceException("Cannot create handle '" + name + "'. Absolute data pointer is null.");
     }
-
   }
 
-  // Dummy bool parameter to differentiate from absolute encoder constructor
+  /**
+   * \param name The name of the actuator
+   * \param pos A pointer to the storage for this actuator's position
+   * \param vel A pointer to the storage for this actuator's velocity
+   * \param eff A pointer to the storage for this actuator's effort (force or torque)
+   * \param torque_sensor A pointer to the storage for this actuator's torque sensor
+   * \param bool Dummy parameter to differentiate from absolute encoder constructor
+   */
   ActuatorStateHandle(const std::string& name, const double* pos, const double* vel, const double* eff,
-                      const double *torque_sensor, bool )
+                      const double* torque_sensor, bool)
     : name_(name), pos_(pos), vel_(vel), eff_(eff), absolute_pos_(0), torque_sensor_(torque_sensor)
   {
     if (!pos)
@@ -108,53 +158,30 @@ public:
     {
       throw HardwareInterfaceException("Cannot create handle '" + name + "'. Effort data pointer is null.");
     }
-    if(!torque_sensor){
+    if (!torque_sensor)
+    {
       throw HardwareInterfaceException("Cannot create handle '" + name + "'. Torque sensor data pointer is null.");
     }
-
-  }
-
-  ActuatorStateHandle(const std::string& name, const double* pos, const double* vel, const double* eff,
-                      const double* absolute_pos, const double *torque_sensor)
-    : name_(name), pos_(pos), vel_(vel), eff_(eff), absolute_pos_(absolute_pos), torque_sensor_(torque_sensor)
-  {
-    if (!pos)
-    {
-      throw HardwareInterfaceException("Cannot create handle '" + name + "'. Position data pointer is null.");
-    }
-    if (!vel)
-    {
-      throw HardwareInterfaceException("Cannot create handle '" + name + "'. Velocity data pointer is null.");
-    }
-    if (!eff)
-    {
-      throw HardwareInterfaceException("Cannot create handle '" + name + "'. Effort data pointer is null.");
-    }
-
-    if (!absolute_pos)
-    {
-      throw HardwareInterfaceException("Cannot create handle '" + name + "'. Absolute data pointer is null.");
-    }
-    if(!torque_sensor){
-      throw HardwareInterfaceException("Cannot create handle '" + name + "'. Torque sensor data pointer is null.");
-    }
-
   }
 
   std::string getName() const {return name_;}
   double getPosition()  const {assert(pos_); return *pos_;}
   double getVelocity()  const {assert(vel_); return *vel_;}
   double getEffort()    const {assert(eff_); return *eff_;}
-  double getAbsolutePosition() const {
-    assert(absolute_pos_);
-    if(!absolute_pos_){
-     throw std::runtime_error("Actuator does not support absolute encoders");
+
+  double getAbsolutePosition() const 
+  {
+    if(!hasAbsolutePosition())
+    {
+      throw std::runtime_error("Actuator does not support absolute encoders");
     }
     return *absolute_pos_;
   }
-  double getTorqueSensor() const {
-    assert(torque_sensor_);
-    if(!torque_sensor_){
+
+  double getTorqueSensor() const 
+  {
+    if(!hasTorqueSensor())
+    {
       throw std::runtime_error("Actuator does not support torque sensors");
     }
     return *torque_sensor_;
@@ -163,36 +190,27 @@ public:
   const double* getPositionPtr() const {return pos_;}
   const double* getVelocityPtr() const {return vel_;}
   const double* getEffortPtr()   const {return eff_;}
-  const double* getAbsolutePositionPtr() const {
-    if(!absolute_pos_){
-     throw std::runtime_error("Actuator does not support absolute encoders");
+
+  const double* getAbsolutePositionPtr() const 
+  {
+    if(!hasAbsolutePosition())
+    {
+      throw std::runtime_error("Actuator does not support absolute encoders");
     }
     return absolute_pos_;
   }
-  const double* getTorqueSensorPtr()   const {
-    if(!torque_sensor_){
+
+  const double* getTorqueSensorPtr() const 
+  {
+    if(!hasTorqueSensor())
+    {
       throw std::runtime_error("Actuator does not support torque sensors");
     }
     return torque_sensor_;
   }
 
-  bool hasAbsolutePosition(){
-    if(!absolute_pos_){
-     return false;
-    }
-    else{
-      return true;
-    }
-  }
-
-  bool hasTorqueSensor(){
-    if(!torque_sensor_){
-     return false;
-    }
-    else{
-      return true;
-    }
-  }
+  bool hasAbsolutePosition()  const {return absolute_pos_;}
+  bool hasTorqueSensor()      const {return torque_sensor_;}
 
 private:
   std::string name_;
--- a/hardware_interface/include/hardware_interface/actuator_state_interface.h
+++ b/hardware_interface/include/hardware_interface/actuator_state_interface.h
@@ -41,7 +41,7 @@ namespace hardware_interface
 class ActuatorStateHandle
 {
 public:
-  ActuatorStateHandle() : name_(), pos_(0), vel_(0), eff_(0) {}
+  ActuatorStateHandle() : name_(), pos_(0), vel_(0), eff_(0), absolute_pos_(0), torque_sensor_(0) {}
 
   /**
    * \param name The name of the actuator
@@ -49,8 +49,9 @@ public:
    * \param vel A pointer to the storage for this actuator's velocity
    * \param eff A pointer to the storage for this actuator's effort (force or torque)
    */
+
   ActuatorStateHandle(const std::string& name, const double* pos, const double* vel, const double* eff)
-    : name_(name), pos_(pos), vel_(vel), eff_(eff)
+    : name_(name), pos_(pos), vel_(vel), eff_(eff), absolute_pos_(0), torque_sensor_(0)
   {
     if (!pos)
     {
@@ -64,22 +65,142 @@ public:
     {
       throw HardwareInterfaceException("Cannot create handle '" + name + "'. Effort data pointer is null.");
     }
+
+  }
+
+  ActuatorStateHandle(const std::string& name, const double* pos, const double* vel, const double* eff,
+                      const double* absolute_pos)
+    : name_(name), pos_(pos), vel_(vel), eff_(eff), absolute_pos_(absolute_pos), torque_sensor_(0)
+  {
+    if (!pos)
+    {
+      throw HardwareInterfaceException("Cannot create handle '" + name + "'. Position data pointer is null.");
+    }
+    if (!vel)
+    {
+      throw HardwareInterfaceException("Cannot create handle '" + name + "'. Velocity data pointer is null.");
+    }
+    if (!eff)
+    {
+      throw HardwareInterfaceException("Cannot create handle '" + name + "'. Effort data pointer is null.");
+    }
+    if (!absolute_pos)
+    {
+      throw HardwareInterfaceException("Cannot create handle '" + name + "'. Absolute data pointer is null.");
+    }
+
+  }
+
+  // Dummy bool parameter to differentiate from absolute encoder constructor
+  ActuatorStateHandle(const std::string& name, const double* pos, const double* vel, const double* eff,
+                      const double *torque_sensor, bool )
+    : name_(name), pos_(pos), vel_(vel), eff_(eff), absolute_pos_(0), torque_sensor_(torque_sensor)
+  {
+    if (!pos)
+    {
+      throw HardwareInterfaceException("Cannot create handle '" + name + "'. Position data pointer is null.");
+    }
+    if (!vel)
+    {
+      throw HardwareInterfaceException("Cannot create handle '" + name + "'. Velocity data pointer is null.");
+    }
+    if (!eff)
+    {
+      throw HardwareInterfaceException("Cannot create handle '" + name + "'. Effort data pointer is null.");
+    }
+    if(!torque_sensor){
+      throw HardwareInterfaceException("Cannot create handle '" + name + "'. Torque sensor data pointer is null.");
+    }
+
+  }
+
+  ActuatorStateHandle(const std::string& name, const double* pos, const double* vel, const double* eff,
+                      const double* absolute_pos, const double *torque_sensor)
+    : name_(name), pos_(pos), vel_(vel), eff_(eff), absolute_pos_(absolute_pos), torque_sensor_(torque_sensor)
+  {
+    if (!pos)
+    {
+      throw HardwareInterfaceException("Cannot create handle '" + name + "'. Position data pointer is null.");
+    }
+    if (!vel)
+    {
+      throw HardwareInterfaceException("Cannot create handle '" + name + "'. Velocity data pointer is null.");
+    }
+    if (!eff)
+    {
+      throw HardwareInterfaceException("Cannot create handle '" + name + "'. Effort data pointer is null.");
+    }
+
+    if (!absolute_pos)
+    {
+      throw HardwareInterfaceException("Cannot create handle '" + name + "'. Absolute data pointer is null.");
+    }
+    if(!torque_sensor){
+      throw HardwareInterfaceException("Cannot create handle '" + name + "'. Torque sensor data pointer is null.");
+    }
+
   }
 
   std::string getName() const {return name_;}
   double getPosition()  const {assert(pos_); return *pos_;}
   double getVelocity()  const {assert(vel_); return *vel_;}
   double getEffort()    const {assert(eff_); return *eff_;}
+  double getAbsolutePosition() const {
+    assert(absolute_pos_);
+    if(!absolute_pos_){
+     throw std::runtime_error("Actuator does not support absolute encoders");
+    }
+    return *absolute_pos_;
+  }
+  double getTorqueSensor() const {
+    assert(torque_sensor_);
+    if(!torque_sensor_){
+      throw std::runtime_error("Actuator does not support torque sensors");
+    }
+    return *torque_sensor_;
+  }
 
   const double* getPositionPtr() const {return pos_;}
   const double* getVelocityPtr() const {return vel_;}
   const double* getEffortPtr()   const {return eff_;}
+  const double* getAbsolutePositionPtr() const {
+    if(!absolute_pos_){
+     throw std::runtime_error("Actuator does not support absolute encoders");
+    }
+    return absolute_pos_;
+  }
+  const double* getTorqueSensorPtr()   const {
+    if(!torque_sensor_){
+      throw std::runtime_error("Actuator does not support torque sensors");
+    }
+    return torque_sensor_;
+  }
+
+  bool hasAbsolutePosition(){
+    if(!absolute_pos_){
+     return false;
+    }
+    else{
+      return true;
+    }
+  }
+
+  bool hasTorqueSensor(){
+    if(!torque_sensor_){
+     return false;
+    }
+    else{
+      return true;
+    }
+  }
 
 private:
   std::string name_;
   const double* pos_;
   const double* vel_;
   const double* eff_;
+  const double* absolute_pos_;
+  const double* torque_sensor_;
 };
 
 /** \brief Hardware interface to support reading the state of an array of actuators
--- a/hardware_interface/include/hardware_interface/joint_command_interface.h
+++ b/hardware_interface/include/hardware_interface/joint_command_interface.h
@@ -59,6 +59,7 @@ public:
 
   void setCommand(double command) {assert(cmd_); *cmd_ = command;}
   double getCommand() const {assert(cmd_); return *cmd_;}
+  const double* getCommandPtr() const {assert(cmd_); return cmd_;}
 
 private:
   double* cmd_;
--- a/hardware_interface/include/hardware_interface/joint_mode_interface.h
+++ b/hardware_interface/include/hardware_interface/joint_mode_interface.h
@@ -61,7 +61,7 @@ public:
   /**
    * \param mode Which mode to start in
    */
-  JointModeHandle(std::string name, int* mode)
+  JointModeHandle(std::string name, JointCommandModes* mode)
     : mode_(mode)
     , name_(name)
   {
@@ -73,11 +73,11 @@ public:
 
   std::string getName() const {return name_;}
 
-  void setMode(int mode) {assert(mode_); *mode_ = mode;}
+  void setMode(JointCommandModes mode) {assert(mode_); *mode_ = mode;}
   int getMode() const {assert(mode_); return *mode_;}
 
   // Helper function for console messages
-  std::string getModeName(int mode)
+  std::string getModeName(JointCommandModes mode)
   {
     switch(mode)
     {
@@ -94,7 +94,7 @@ public:
   }
 
 private:
-  int* mode_;
+  JointCommandModes* mode_;
   std::string name_;
 };
 
--- a/hardware_interface/include/hardware_interface/joint_mode_interface.h
+++ b/hardware_interface/include/hardware_interface/joint_mode_interface.h
@@ -62,8 +62,8 @@ public:
    * \param mode Which mode to start in
    */
   JointModeHandle(std::string name, int* mode)
-    : name_(name),
-      mode_(mode)
+    : mode_(mode)
+    , name_(name)
   {
     if (!mode_)
     {
--- a/hardware_interface/include/hardware_interface/joint_mode_interface.h
+++ b/hardware_interface/include/hardware_interface/joint_mode_interface.h
@@ -1,0 +1,108 @@
+/*********************************************************************
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2013, University of Colorado, Boulder
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions
+ *  are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above
+ *     copyright notice, this list of conditions and the following
+ *     disclaimer in the documentation and/or other materials provided
+ *     with the distribution.
+ *   * Neither the name of the Univ of CO, Boulder nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ *  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ *  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ *  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ *  COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ *  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ *  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
+
+/* Author: Dave Coleman
+   Desc:   This interface is for switching a hardware interface between different controller modes
+           i.e. position, velocity, force
+*/
+
+#ifndef HARDWARE_INTERFACE_JOINT_MODE_INTERFACE_H
+#define HARDWARE_INTERFACE_JOINT_MODE_INTERFACE_H
+
+#include <cassert>
+#include <hardware_interface/internal/hardware_resource_manager.h>
+
+namespace hardware_interface
+{
+
+enum JointCommandModes {
+  MODE_POSITION = 1,
+  MODE_VELOCITY = 2,
+  MODE_EFFORT = 3,
+  MODE_OTHER = 4
+};
+
+/** \brief A handle used to read and mode a single joint. */
+class JointModeHandle
+{
+public:
+
+  /**
+   * \param mode Which mode to start in
+   */
+  JointModeHandle(std::string name, int* mode)
+    : name_(name),
+      mode_(mode)
+  {
+    if (!mode_)
+    {
+      throw HardwareInterfaceException("Cannot create mode interface. Mode data pointer is null.");
+    }
+  }
+
+  std::string getName() const {return name_;}
+
+  void setMode(int mode) {assert(mode_); *mode_ = mode;}
+  int getMode() const {assert(mode_); return *mode_;}
+
+  // Helper function for console messages
+  std::string getModeName(int mode)
+  {
+    switch(mode)
+    {
+      case MODE_POSITION:
+        return "position";
+      case MODE_VELOCITY:
+        return "velocity";
+      case MODE_EFFORT:
+        return "effort";
+      case MODE_OTHER:
+        return "other";
+    }
+    return "unknown";
+  }
+
+private:
+  int* mode_;
+  std::string name_;
+};
+
+/** \brief Hardware interface to support changing between control modes
+ *
+ */
+class JointModeInterface : public HardwareResourceManager<JointModeHandle, ClaimResources> {};
+
+} // namespace
+
+#endif
--- a/hardware_interface/include/hardware_interface/joint_mode_interface.h
+++ b/hardware_interface/include/hardware_interface/joint_mode_interface.h
@@ -34,11 +34,10 @@
 
 /* Author: Dave Coleman
    Desc:   This interface is for switching a hardware interface between different controller modes
-           i.e. position, velocity, force
+           i.e. position, velocity, effort
 */
 
-#ifndef HARDWARE_INTERFACE_JOINT_MODE_INTERFACE_H
-#define HARDWARE_INTERFACE_JOINT_MODE_INTERFACE_H
+#pragma once
 
 #include <cassert>
 #include <hardware_interface/internal/hardware_resource_manager.h>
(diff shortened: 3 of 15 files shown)


hardware_interface/include/hardware_interface/actuator_state_interface.h
```diff
@@ -64,22 +65,142 @@ class ActuatorStateHandle
     {
       throw HardwareInterfaceException("Cannot create handle '" + name + "'. Effort data pointer is null.");
     }
+
```
matthew-reynolds: Unneccesary blank line before closing brace. This occurs in several new functions.

hardware_interface/include/hardware_interface/actuator_state_interface.h
```diff
@@ -64,22 +65,142 @@ class ActuatorStateHandle
...
 
   std::string getName() const {return name_;}
   double getPosition()  const {assert(pos_); return *pos_;}
   double getVelocity()  const {assert(vel_); return *vel_;}
   double getEffort()    const {assert(eff_); return *eff_;}
+  double getAbsolutePosition() const {
+    assert(absolute_pos_);
+    if(!absolute_pos_){
+     throw std::runtime_error("Actuator does not support absolute encoders");
+    }
```
matthew-reynolds: `assert` should be removed, as the `absolute_pos_` ptr is optional. Same below for `torque_sensor_`.

hardware_interface/include/hardware_interface/actuator_state_interface.h
```diff
@@ -64,22 +65,142 @@ class ActuatorStateHandle
...
+  }
+
+  bool hasAbsolutePosition(){
+    if(!absolute_pos_){
+     return false;
+    }
+    else{
+      return true;
+    }
+  }
```
matthew-reynolds: Should be const, and simplified as:

```
bool hasAbsolutePosition() const {
    return absolute_pos_;
}
```

Same below for `hasTorqueSensor()`.

hardware_interface/include/hardware_interface/actuator_state_interface.h
```diff
@@ -64,22 +65,142 @@ class ActuatorStateHandle
...
+    return absolute_pos_;
+  }
+  const double* getTorqueSensorPtr()   const {
+    if(!torque_sensor_){
+      throw std::runtime_error("Actuator does not support torque sensors");
+    }
+    return torque_sensor_;
+  }
+
+  bool hasAbsolutePosition(){
```
matthew-reynolds: Should consider using these new funcs in `getAbsolutePosition()`, `getAbsolutePositionPtr()`, `getTorqueSensor()`, and `getTorqueSensorPtr()` to be consistent with `joint_state_interface.h`.

hardware_interface/include/hardware_interface/internal/resource_manager.h
```diff
@@ -118,6 +118,11 @@ class ResourceManager : public ResourceManagerBase
 
     if (it == resource_map_.end())
     {
+      ROS_DEBUG_STREAM_NAMED("resource_manager","Available resource handles:");
```
matthew-reynolds: If this debug print is to remain, consider printing the same text as the `logic_error` ("Could not find resource 'name' in '*this'") before listing the available resource handles

hardware_interface/include/hardware_interface/joint_command_interface.h
```diff
@@ -59,6 +59,7 @@ class JointHandle : public JointStateHandle
 
   void setCommand(double command) {assert(cmd_); *cmd_ = command;}
   double getCommand() const {assert(cmd_); return *cmd_;}
+  const double* getCommandPtr() const {assert(cmd_); return cmd_;}
```
matthew-reynolds, line 62: What is the rational for mimicking the `actuator_command_interface` and `actuator_state_interface` (#135) by exposing the cmd and state ptrs in `joint_state/command_interface`?

Also, `ActuatorHandle::getCommandPtr()` returns `double*`, not `const double*` (though all the uses of `ActuatorHandle::getCommandPtr()` are redundantly `const_cast`ed to `double*` anyways). I'm unfamiliar with the intentions of this accessor, but it seems like either `const` should be removed here or more likely added there.

hardware_interface/include/hardware_interface/joint_mode_interface.h
```diff
@@ -0,0 +1,124 @@
...
+  {
+  public:
+
+    JointModeHandle():
+      name_(), mode_(0){}
+
+    /**
+   * \param mode Which mode to start in
+   */
+    JointModeHandle(std::string name, int* mode)
```
matthew-reynolds: Throughout this file, use `JointCommandModes*` rather than `int*`.

It looks like this change was already made in d9d8c57, but was accidentally reverted in 8750f0bb.

hardware_interface/include/hardware_interface/joint_mode_interface.h
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    std::string getName() const {return name_;}
+
+    // void setMode(JointCommandModes mode) {assert(mode_); *mode_ = mode;}
+    int getMode() const {assert(mode_); return *mode_;}
+    const int* getModePtr() const {assert(mode_); return mode_;}
+
+    // Helper function for console messages
+    std::string getModeName(JointCommandModes mode)
```
matthew-reynolds: Should be const

hardware_interface/include/hardware_interface/joint_mode_interface.h
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+  private:
+    int* mode_;
+    std::string name_;
+  };
+
+  /** \brief Hardware interface to support changing between control modes
+ *
+ */
```
matthew-reynolds: Comment should be collapsed to single line

`/** \brief Hardware interface to support changing between control modes */`

hardware_interface/include/hardware_interface/actuator_state_interface.h
```diff
@@ -41,16 +41,17 @@ namespace hardware_interface
...
-  ActuatorStateHandle() : name_(), pos_(0), vel_(0), eff_(0) {}
+  ActuatorStateHandle() : name_(), pos_(0), vel_(0), eff_(0), absolute_pos_(0), torque_sensor_(0) {}
 
   /**
    * \param name The name of the actuator
    * \param pos A pointer to the storage for this actuator's position
    * \param vel A pointer to the storage for this actuator's velocity
    * \param eff A pointer to the storage for this actuator's effort (force or torque)
    */
+
```
matthew-reynolds: Extra newline. Also, function comment should be updated to include new parameters.

Same applies to `joint_state_interface.h`

hardware_interface/include/hardware_interface/actuator_state_interface.h
```diff
@@ -64,22 +65,142 @@ class ActuatorStateHandle
     {
       throw HardwareInterfaceException("Cannot create handle '" + name + "'. Effort data pointer is null.");
     }
+
+  }
+
+  ActuatorStateHandle(const std::string& name, const double* pos, const double* vel, const double* eff,
```
matthew-reynolds, line 81: New constructors missing function comments. In addition to parameter descriptions, should consider adding brief blurb highlighting the differences between the constructors.

Same applies to `joint_state_interface.h`

hardware_interface/include/hardware_interface/joint_state_interface.h
```diff
@@ -41,16 +41,92 @@ namespace hardware_interface
...
    * \param pos A pointer to the storage for this joint's position
    * \param vel A pointer to the storage for this joint's velocity
    * \param eff A pointer to the storage for this joint's effort (force or torque)
    */
+
+  JointStateHandle(const std::string& name, const double* pos, const double* vel, const double* eff,
+                   const double* absolute_pos, const double *torque_sensor)
+    : name_(name), pos_(pos), vel_(vel), eff_(eff), absolute_pos_(absolute_pos), torque_sensor_(torque_sensor)
+  {
+
```
matthew-reynolds: Extra newline at beginning of all these constructors.

hardware_interface/include/hardware_interface/joint_state_interface.h
```diff
@@ -41,16 +41,92 @@ namespace hardware_interface
...
-  JointStateHandle() : name_(), pos_(0), vel_(0), eff_(0) {}
+  JointStateHandle() : name_(), pos_(0), vel_(0), eff_(0), absolute_pos_(0), torque_sensor_(0) {}
 
   /**
    * \param name The name of the joint
    * \param pos A pointer to the storage for this joint's position
    * \param vel A pointer to the storage for this joint's velocity
    * \param eff A pointer to the storage for this joint's effort (force or torque)
    */
+
```
matthew-reynolds, line 72: Extra newline. Also, see comment above about the function comment on the constructors in this file.

hardware_interface/include/hardware_interface/joint_state_interface.h
```diff
@@ -64,18 +140,75 @@ class JointStateHandle
     {
       throw HardwareInterfaceException("Cannot create handle '" + name + "'. Effort data pointer is null.");
     }
+
```
matthew-reynolds: Extra newline

hardware_interface/include/hardware_interface/joint_state_interface.h
```diff
@@ -64,18 +140,75 @@ class JointStateHandle
...
 
+  const double* getPositionPtr() const {return pos_;}
+  const double* getVelocityPtr() const {return vel_;}
+  const double* getEffortPtr()   const {return eff_;}
+
+  double getAbsolutePosition() const {
+    assert(absolute_pos_);
+    if(!hasAbsolutePosition()){
+      throw std::runtime_error("Joint state handle does not contain absolute encoder position information");
+    }
```
matthew-reynolds: As in `actuator_state_interface.h`, these `assert`s should be removed

hardware_interface/include/hardware_interface/joint_state_interface.h
```diff
@@ -64,18 +140,75 @@ class JointStateHandle
...
   }
 
   std::string getName() const {return name_;}
   double getPosition()  const {assert(pos_); return *pos_;}
   double getVelocity()  const {assert(vel_); return *vel_;}
   double getEffort()    const {assert(eff_); return *eff_;}
 
+  const double* getPositionPtr() const {return pos_;}
+  const double* getVelocityPtr() const {return vel_;}
+  const double* getEffortPtr()   const {return eff_;}
```
matthew-reynolds: In `actuator_state_interface.h`, all the data getters are together, followed by all the ptr getters. This file should follow suit.

hardware_interface/include/hardware_interface/joint_state_interface.h
```diff
@@ -64,18 +140,75 @@ class JointStateHandle
...
+    }
+    return torque_sensor_;
+  }
+
+  bool hasAbsolutePosition() const {
+    if(absolute_pos_){
+      return true;
+    }
+    return false;
+  }
```
matthew-reynolds: As in `actuator_state_interface`, simplify these two funcs to single-line `return absolute_pos_;`/`return torque_sensor_;`.

hardware_interface/include/hardware_interface/joint_state_interface.h
```diff
@@ -64,18 +140,75 @@ class JointStateHandle
...
+      return false;
+    }
+  }
+
 private:
   std::string name_;
   const double* pos_;
   const double* vel_;
   const double* eff_;
+
```
matthew-reynolds: Extra newline

joint_limits_interface/test/joint_limits_interface_test.cpp
```diff
@@ -376,28 +376,41 @@ TEST_F(VelocityJointSaturationHandleTest, EnforceAccelerationBounds)
 
   pos = 0.0;
   double cmd;
+  ros::Duration long_enough(1000.0);
```
matthew-reynolds: Should be const, and should include comment of why 1000sec was selected, even if it's just `// An arbitrarily long time, sufficient to suppress acceleration limits`,

transmission_interface/include/transmission_interface/differential_transmission.h
```diff
@@ -264,6 +281,38 @@ inline void DifferentialTransmission::actuatorToJointPosition(const ActuatorData
...
 }
 
+inline void DifferentialTransmission::actuatorToJointAbsolutePosition(const ActuatorData& act_data,
+                                                                    JointData&    jnt_data)
+{
+  assert(numActuators() == act_data.absolute_position.size() && numJoints() == jnt_data.absolute_position.size());
+  assert(act_data.position[0] && act_data.absolute_position[1] && jnt_data.absolute_position[0] && jnt_data.absolute_position[1]);
+
+  std::vector<double>& ar = act_reduction_;
+  std::vector<double>& jr = jnt_reduction_;
```
matthew-reynolds: Should be at least `const`. Not sure what this project's preferences are on things like `auto`, but could consider `const auto&`.

transmission_interface/include/transmission_interface/differential_transmission.h
```diff
@@ -123,7 +123,8 @@ class DifferentialTransmission : public Transmission
    * \param joint_offset       Joint position offset used in the position mappings.
    * \pre Nonzero actuator and joint reduction values.
    */
-  DifferentialTransmission(const std::vector<double>& actuator_reduction,
+  DifferentialTransmission(const bool &ignore_transmission_for_absolute_encoders,
```
matthew-reynolds: Why is this bool passed as const ref? Should be pass by val.

transmission_interface/include/transmission_interface/differential_transmission.h
```diff
@@ -157,6 +158,19 @@ class DifferentialTransmission : public Transmission
   void actuatorToJointPosition(const ActuatorData& act_data,
                                      JointData&    jnt_data);
 
+  void actuatorToJointAbsolutePosition(const ActuatorData& act_data,
+                                             JointData&    jnt_data);
+
+  void actuatorToJointTorqueSensor(const ActuatorData& act_data,
+                                             JointData&    jnt_data);
```
matthew-reynolds: Missing function comments.

transmission_interface/include/transmission_interface/differential_transmission.h
matthew-reynolds: Just realized that these are simply mimicking the pattern in the exsiting functions. They should be const both here and there, though perhaps that's better as a follow-up PR

transmission_interface/include/transmission_interface/transmission.h
```diff
@@ -122,6 +126,20 @@ class Transmission
...
+  }
+
+  virtual void actuatorToJointTorqueSensor(const ActuatorData&,
+                                             JointData&){
+    throw std::runtime_error("transmission does not support actuator to joint torque sensor");
+  }
+
+  virtual bool hasActuatorToJointAbsolutePosition() = 0;
+
+  virtual bool hasActuatorToJointTorqueSensor() = 0;
```
matthew-reynolds: Should be const

transmission_interface/include/transmission_interface/transmission.h
```diff
@@ -122,6 +126,20 @@ class Transmission
...
 
+  virtual void actuatorToJointAbsolutePosition(const ActuatorData&,
+                                             JointData&){
+    throw std::runtime_error("transmission does not support actuator to joint absolute position");
+  }
+
+  virtual void actuatorToJointTorqueSensor(const ActuatorData&,
+                                             JointData&){
+    throw std::runtime_error("transmission does not support actuator to joint torque sensor");
+  }
```
matthew-reynolds: These should be const, but I know that differs from all the other actuatorToJointX() and jointToActuatorX() functions. Is there a reason the existing functions aren't const?

transmission_interface/include/transmission_interface/transmission_interface_loader.h
```diff
@@ -119,6 +123,12 @@ struct RawJointData
   double position_cmd;
   double velocity_cmd;
   double effort_cmd;
+
```
matthew-reynolds: Extra newline

transmission_interface/src/differential_transmission_loader.cpp
```diff
@@ -239,6 +243,17 @@ bool DifferentialTransmissionLoader::getJointConfig(const TransmissionInfo& tran
...
 
+  // Parse if transmision has to be ignored for absolute encoders
+  ignore_transmission_for_absolute_encoders = true;
+  for (unsigned int i = 0; i < 2; ++i){
+    const TiXmlElement* role_el = jnt_elements[i].FirstChildElement("ignoreTransmissionAbsoluteEncoder");
+    bool ignore = false;
+    if(role_el){
+      ignore = true;
+    }
+    ignore_transmission_for_absolute_encoders &= ignore;
```
matthew-reynolds: This is redundant. Can just be:

`ignore_transmission_for_absolute_encoders &=static_cast<bool>(role_el);` or `(role_el != nullptr)` etc.

matthew-reynolds (commented):
These changes should all be run through an auto-formatter. There are a lot of stylistic differences between the new changes and the rest of the codebase.

I realized while going through some of the comments I had were about established patterns. This is already a fairly large PR, so perhaps it makes more sense to stick with the established patterns and those changes should be pulled out into follow-ups (ie. replacing `0` with `nullptr`, adding `consts` all over the place, etc.)

hardware_interface/include/hardware_interface/joint_command_interface.h
matthew-reynolds: transmission_interface/src/joint_state_interface_provider.cpp ln 188:
`// TODO: Get rid of these const casts!`

Looks like I need to read a bit more of the code to get a better understanding of this whole `const double*` situation in both the `actuator_state/command_interface.h` and `joint_state/command_interface.h`.

hardware_interface/include/hardware_interface/actuator_state_interface.h
bmagyar: very good catch, fixed and thanks!

hardware_interface/include/hardware_interface/actuator_state_interface.h
bmagyar: indeed, this is actually used on job interviews ;)

hardware_interface/include/hardware_interface/actuator_state_interface.h
bmagyar: neat, looks much better

hardware_interface/include/hardware_interface/joint_mode_interface.h
bmagyar: I think the story of this horrible section involves several autoformatters that did their worst on it :)

hardware_interface/include/hardware_interface/actuator_state_interface.h
bmagyar: righty'ho!

hardware_interface/include/hardware_interface/joint_state_interface.h
bmagyar: too much copy pasta

hardware_interface/include/hardware_interface/joint_state_interface.h
bmagyar: thanks

hardware_interface/include/hardware_interface/joint_state_interface.h
bmagyar: I guess this was artistic freedom ;)

transmission_interface/include/transmission_interface/transmission.h
bmagyar: indeed!

transmission_interface/src/differential_transmission_loader.cpp
bmagyar: it feels like I've fixed this a 100 times already, thanks

hardware_interface/include/hardware_interface/internal/resource_manager.h
bmagyar: we'll rewrite it when needed, won't add to upstream

hardware_interface/include/hardware_interface/joint_mode_interface.h
bmagyar: :+1:

joint_limits_interface/test/joint_limits_interface_test.cpp
bmagyar: awesome

transmission_interface/include/transmission_interface/differential_transmission.h
bmagyar: a `const` val is has a tiny tiny caveat that it doesn't even accidentally allow you to change the value received as an argument, I settled for that.

hardware_interface/include/hardware_interface/actuator_state_interface.h
bmagyar: done

hardware_interface/include/hardware_interface/joint_mode_interface.h
bmagyar: I undid the undoing

hardware_interface/include/hardware_interface/joint_state_interface.h
bmagyar: it should now

transmission_interface/include/transmission_interface/differential_transmission.h
bmagyar: added

transmission_interface/include/transmission_interface/differential_transmission.h
matthew-reynolds: Fair enough, though per the [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html#Use_of_const), `const` should not be used on parameters passed by value in function _declarations_, as it is meaningless. Using it in _definitions_ is neither encouraged nor discouraged, but left to the team/project's discretion, as they are equivalent to local const variables.

I agree with you that we should move toward adding `const` for local vars and, equivalently, function parameters passed by value in _definitions_, but not declarations. 

A final note: Adding `const` here is inconsistent with the style of the rest of the project, but since one of the topics in #403 is adding `const` where appropriate, in time the rest of the project will be brought in line with this convention :)

transmission_interface/include/transmission_interface/transmission.h
bmagyar: We should allow internal states to be used&altered if needed in these functions

transmission_interface/include/transmission_interface/differential_transmission.h
bmagyar: I've added consts to the other functions too

transmission_interface/include/transmission_interface/transmission.h
matthew-reynolds: What internal state would/could be modified? Currently, the 3 transmissions implemented (differential, four bar, and simple) do not modify any internal state, and I can't think of an example where you'd want a data transformation function to have side effects.

hardware_interface/include/hardware_interface/joint_command_interface.h
bmagyar: I agree with the sentiment but would prefer to solve this in a separate PR

transmission_interface/include/transmission_interface/transmission.h
bmagyar: Transmissions are plugin-based, anyone can write their own with whatever internal logic they wish to use. Of course these are niche cases but my point is, there's a possibility of a use-case we shouldn't rule out proactively.

bmagyar:
@matthew-reynolds could you please give it another pass? I think it should be ready now

hardware_interface/include/hardware_interface/joint_command_interface.h
matthew-reynolds: :+1: I'll get a better understanding and follow up

hardware_interface/include/hardware_interface/joint_mode_interface.h
```diff
@@ -0,0 +1,118 @@
...
+
+#pragma once
+
+#include <cassert>
+#include <hardware_interface/internal/hardware_resource_manager.h>
+
+namespace hardware_interface
+{
+
+  enum class JointCommandModes : int
```
matthew-reynolds: No harm in specifying the underlying type as `int`, but it isn't necessary as `int` is the default. If we're going to bother specifying one at all, is there a reason to prefer `int` over something more compact like `signed char` or maybe `int8_t`?

hardware_interface/include/hardware_interface/joint_state_interface.h
```diff
@@ -37,20 +37,113 @@
...
 public:
-  JointStateHandle() : name_(), pos_(0), vel_(0), eff_(0) {}
+  JointStateHandle() : name_(), pos_(0), vel_(0), eff_(0), absolute_pos_(0), torque_sensor_(0) {}
+
+  /**
+   * \param name The name of the actuator
+   * \param pos A pointer to the storage for this actuator's position
+   * \param vel A pointer to the storage for this actuator's velocity
+   * \param eff A pointer to the storage for this actuator's effort (force or torque)
+   */
```
matthew-reynolds: Function comment should use "joint", not "actuator". Same goes for several other constructors as well.

transmission_interface/include/transmission_interface/differential_transmission_loader.h
```diff
@@ -53,7 +53,8 @@ class DifferentialTransmissionLoader : public TransmissionLoader
 
   static bool getJointConfig(const TransmissionInfo& transmission_info,
                              std::vector<double>&    joint_reduction,
-                             std::vector<double>&    joint_offset);
+                             std::vector<double>&    joint_offset,
+                             bool &ignore_transmission_for_absolute_encoders);
```
matthew-reynolds: Parameter `ignore_transmission_for_absolute_encoders` should be aligned with other parameters, and `&` should be on the left.

transmission_interface/include/transmission_interface/differential_transmission.h
```diff
@@ -157,6 +158,25 @@ class DifferentialTransmission : public Transmission
...
+   */
+  void actuatorToJointAbsolutePosition(const ActuatorData& act_data,
+                                             JointData&    jnt_data);
+
+  /**
+   * \brief Transform \e torque sensor values from actuator to joint space.
+   * \param[in]  act_data Actuator-space variables.
+   * \param[out] jnt_data Joint-space variables.
+   * \pre Actuator, joint position and torque sensor vectors must have the same size.
+   */
```
matthew-reynolds: Could be missing something, but is there a reason we don't have the same "To call this method it is not required that all other data vectors contain valid data, and can even remain empty" blurb here?

transmission_interface/include/transmission_interface/differential_transmission.h
```diff
@@ -123,7 +123,8 @@ class DifferentialTransmission : public Transmission
    * \param joint_offset       Joint position offset used in the position mappings.
    * \pre Nonzero actuator and joint reduction values.
    */
-  DifferentialTransmission(const std::vector<double>& actuator_reduction,
+  DifferentialTransmission(const bool ignore_transmission_for_absolute_encoders,
```
matthew-reynolds: As mentioned in https://github.com/ros-controls/ros_control/pull/395#discussion_r361810399, `const` here in the declaration is meaningless. It should definitely be present in the definition below, though.

hardware_interface/include/hardware_interface/joint_state_interface.h
```diff
@@ -37,20 +37,113 @@
...
    * \param pos A pointer to the storage for this joint's position
    * \param vel A pointer to the storage for this joint's velocity
    * \param eff A pointer to the storage for this joint's effort (force or torque)
+   * \param absolute_pos A pointer to the storage for this joint's absolute position encoder
+   * \param torque_sensor A pointer to the storage for this joint's torque sensor
    */
-  JointStateHandle(const std::string& name, const double* pos, const double* vel, const double* eff)
-    : name_(name), pos_(pos), vel_(vel), eff_(eff)
+  JointStateHandle(const std::string& name, const double* pos, const double* vel, const double* eff,
+                   const double* absolute_pos, const double *torque_sensor)
```
matthew-reynolds: `*` in `const double *torque_sensor` should be left-aligned.

hardware_interface/include/hardware_interface/joint_state_interface.h
```diff
@@ -37,20 +37,113 @@
...
+  /**
+   * \param name The name of the actuator
+   * \param pos A pointer to the storage for this actuator's position
+   * \param vel A pointer to the storage for this actuator's velocity
+   * \param eff A pointer to the storage for this actuator's effort (force or torque)
+   * \param torque_sensor A pointer to the storage for this actuator's torque sensor
+   * \param bool Dummy parameter to differentiate from absolute encoder constructor
+   */
+  JointStateHandle(const std::string& name, const double* pos, const double* vel, const double* eff,
+                   const double *torque_sensor, bool)
```
matthew-reynolds: `*` in `const double *torque_sensor` should be left-aligned.

hardware_interface/include/hardware_interface/joint_mode_interface.h
```diff
@@ -0,0 +1,118 @@
...
+  {
+  public:
+
+    JointModeHandle():
+      name_(), mode_(0){}
+
+    /** \param mode Which mode to start in */
+    JointModeHandle(std::string name, JointCommandModes* mode)
+      : mode_(mode)
+      , name_(name)
```
matthew-reynolds, line 71: Not a big deal, but the rest of the project seems to keep these on the same line rather than one-per-line.

hardware_interface/include/hardware_interface/actuator_state_interface.h
```diff
@@ -50,7 +54,67 @@ class ActuatorStateHandle
...
+  /**
+   * \param name The name of the actuator
+   * \param pos A pointer to the storage for this actuator's position
+   * \param vel A pointer to the storage for this actuator's velocity
+   * \param eff A pointer to the storage for this actuator's effort (force or torque)
+   * \param absolute_pos A pointer to the storage for this actuator's absolute position encoder
+   * \param torque_sensor A pointer to the storage for this actuator's torque sensor
+   */
+  ActuatorStateHandle(const std::string& name, const double* pos, const double* vel, const double* eff,
+                      const double* absolute_pos, const double *torque_sensor)
```
matthew-reynolds: `*` in `const double *torque_sensor` should be left-aligned.

hardware_interface/include/hardware_interface/actuator_state_interface.h
```diff
@@ -64,22 +128,97 @@ class ActuatorStateHandle
...
+  /**
+   * \param name The name of the actuator
+   * \param pos A pointer to the storage for this actuator's position
+   * \param vel A pointer to the storage for this actuator's velocity
+   * \param eff A pointer to the storage for this actuator's effort (force or torque)
+   * \param torque_sensor A pointer to the storage for this actuator's torque sensor
+   * \param bool Dummy parameter to differentiate from absolute encoder constructor
+   */
+  ActuatorStateHandle(const std::string& name, const double* pos, const double* vel, const double* eff,
+                      const double *torque_sensor, bool)
```
matthew-reynolds: `*` in `const double *torque_sensor` should be left-aligned.

transmission_interface/include/transmission_interface/four_bar_linkage_transmission.h
```diff
@@ -156,6 +156,25 @@ class FourBarLinkageTransmission : public Transmission
...
+                                             JointData&    jnt_data);
+
+  /**
+   * \brief Transform \e torque sensor values from actuator to joint space.
+   * \param[in]  act_data Actuator-space variables.
+   * \param[out] jnt_data Joint-space variables.
+   * \pre Actuator, joint position and torque sensor vectors must have the same size.
+   */
+  void actuatorToJointTorqueSensor(const ActuatorData& act_data,
+                                             JointData&    jnt_data);
```
matthew-reynolds: Weird alignment here

transmission_interface/include/transmission_interface/four_bar_linkage_transmission.h
```diff
@@ -189,6 +208,8 @@ class FourBarLinkageTransmission : public Transmission
 
   std::size_t numActuators() const {return 2;}
   std::size_t numJoints()    const {return 2;}
+  bool hasActuatorToJointAbsolutePosition() const {return true;}
+  bool hasActuatorToJointTorqueSensor() const {return true;}
```
matthew-reynolds: `const`s should be aligned

transmission_interface/include/transmission_interface/four_bar_linkage_transmission.h
```diff
@@ -263,6 +284,36 @@ inline void FourBarLinkageTransmission::actuatorToJointPosition(const ActuatorDa
...
 
+void FourBarLinkageTransmission::actuatorToJointAbsolutePosition(const ActuatorData& act_data,
+                                                                       JointData&    jnt_data)
+{
+
+  assert(numActuators() == act_data.absolute_position.size() && numJoints() == jnt_data.absolute_position.size());
+  assert(act_data.absolute_position[0] && act_data.absolute_position[1] && jnt_data.absolute_position[0] && jnt_data.absolute_position[1]);
+
+  std::vector<double>& ar = act_reduction_;
+  std::vector<double>& jr = jnt_reduction_;
```
matthew-reynolds: As in `differential_transmission.h`, should be `const` throughout these functions

transmission_interface/include/transmission_interface/differential_transmission.h
```diff
@@ -189,6 +209,9 @@ class DifferentialTransmission : public Transmission
 
   std::size_t numActuators() const {return 2;}
   std::size_t numJoints()    const {return 2;}
+  bool hasActuatorToJointAbsolutePosition() const {return true;}
+  bool hasActuatorToJointTorqueSensor() const {return true;}
```
matthew-reynolds: `const`s should be aligned

transmission_interface/include/transmission_interface/simple_transmission.h
```diff
@@ -133,6 +133,12 @@ class SimpleTransmission : public Transmission
   void actuatorToJointPosition(const ActuatorData& act_data,
                                      JointData&    jnt_data);
 
+  void actuatorToJointAbsolutePosition(const ActuatorData& act_data,
+                                             JointData&    jnt_data);
+
+  void actuatorToJointTorqueSensor(const ActuatorData& act_data,
+                                             JointData&    jnt_data);
```
matthew-reynolds: Weird alignment here

transmission_interface/include/transmission_interface/simple_transmission.h
```diff
@@ -165,10 +171,13 @@ class SimpleTransmission : public Transmission
 
   std::size_t numActuators() const {return 1;}
   std::size_t numJoints()    const {return 1;}
+  bool hasActuatorToJointAbsolutePosition() const {return true;}
+  bool hasActuatorToJointTorqueSensor() const {return true;}
```
matthew-reynolds: `const`s should be aligned

transmission_interface/include/transmission_interface/transmission.h
```diff
@@ -122,6 +126,21 @@ class Transmission
...
+
+  virtual void actuatorToJointTorqueSensor(const ActuatorData&,
+                                                 JointData&)
+  {
+    throw std::runtime_error("transmission does not support actuator to joint torque sensor");
+  }
+
+  virtual bool hasActuatorToJointAbsolutePosition() const {return false;}
+  virtual bool hasActuatorToJointTorqueSensor() const {return false;}
```
matthew-reynolds: Consider aligning `const`s here

transmission_interface/src/differential_transmission_loader.cpp
```diff
@@ -154,7 +157,8 @@ bool DifferentialTransmissionLoader::getActuatorConfig(const TransmissionInfo& t
 
 bool DifferentialTransmissionLoader::getJointConfig(const TransmissionInfo& transmission_info,
                                                     std::vector<double>&    joint_reduction,
-                                                    std::vector<double>&    joint_offset)
+                                                    std::vector<double>&    joint_offset,
+                                                    bool &ignore_transmission_for_absolute_encoders)
```
matthew-reynolds: `&` in `bool &ignore_transmission[...]` should be left-aligned

hardware_interface/include/hardware_interface/joint_mode_interface.h
```diff
@@ -0,0 +1,118 @@
...
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
+
+/* Author: Dave Coleman
+   Desc:   This interface is for switching a hardware interface between different controller modes
+           i.e. position, velocity, force
```
matthew-reynolds: Consider "effort" instead of "force" for consistent terminology

transmission_interface/include/transmission_interface/differential_transmission.h
```diff
@@ -123,7 +123,8 @@ class DifferentialTransmission : public Transmission
    * \param joint_offset       Joint position offset used in the position mappings.
    * \pre Nonzero actuator and joint reduction values.
    */
-  DifferentialTransmission(const std::vector<double>& actuator_reduction,
+  DifferentialTransmission(const bool ignore_transmission_for_absolute_encoders,
+                           const std::vector<double>& actuator_reduction,
                            const std::vector<double>& joint_reduction,
                            const std::vector<double>& joint_offset = std::vector<double>(2, 0.0));
```
matthew-reynolds: Looking at this again, if this is targetting Melodic then one of the two changes should be made to avoid breaking changes:

 - Using delegating constructors, add another constructor with the old signature, and a default value `false` for `ignore_transmission_for_absolute_encoders`
 - Make `ignore_transmission_for_absolute_encoders` the last parameter, with default value `false`

matthew-reynolds (commented):
There's a breaking change in the DifferentialTransmission constructor that should be fixed up. Once that's fixed up, there are a couple of style nitpicks but LGTM.

Beyond the code itself, I've got a couple questions about the usages and need for these changes, but I suspect you've already thought them through and I'll understand better once some more documentation is present. Don't want to hold up the PR while I'm getting conceptually on the same page, since the code itself looks good.

We should definitely get some documentation going that explains:
 - `effort` vs `torque_sensor` and `pos` vs `absolute_position`
 - The usage of the `JointModeHandle` (Once it's fleshed out)
 - What `ignore_transmission_for_absolute_encoders` does in `DifferentialTransmission` and how the loader can read it

Also, for changelogs down the road, it should be pointed out that this mega-PR:
 - Adds `absolute_position` and `torque_sensor` data types to JointStateHandle and ActuatorStateHandle, as well as the corresponding transformations to Transmissions
 - Changes the way velocity joint saturation limits are calculated (Now uses prev commanded vel rather than measured vel).
 - Exposes raw data pointers in JointStateHandle and JointCommandHandle (Previously only in actuator handles)
 - Adds a new JointModeHandle "for switching a hardware interface between different controller modes"

(Might be worth updating the PR description so that people reading this in the future don't need to go digging through comments)


Also, a final comment: Why `torque_sensor` rather than `torque`?

hardware_interface/include/hardware_interface/joint_mode_interface.h
bmagyar: You are right. Let's stick to the default for now until and perhaps open a discussion on the actual value type later.

hardware_interface/include/hardware_interface/joint_state_interface.h
bmagyar: good spotting! thanks

transmission_interface/include/transmission_interface/differential_transmission_loader.h
bmagyar: :+1:

transmission_interface/include/transmission_interface/differential_transmission.h
bmagyar: :+1:

transmission_interface/src/differential_transmission_loader.cpp
bmagyar: :+1:

hardware_interface/include/hardware_interface/joint_mode_interface.h
bmagyar: good catch!

transmission_interface/src/differential_transmission_loader.cpp
matthew-reynolds: You probably already caught it, but the parameter name should be aligned with the others, forgot to write that. Same as https://github.com/ros-controls/ros_control/pull/395#discussion_r361883702

transmission_interface/include/transmission_interface/transmission.h
bmagyar: aye!

transmission_interface/include/transmission_interface/simple_transmission.h
bmagyar: right!

transmission_interface/include/transmission_interface/differential_transmission_loader.h
bmagyar: thanks

transmission_interface/include/transmission_interface/differential_transmission.h
bmagyar: gone

transmission_interface/src/differential_transmission_loader.cpp
bmagyar: absolutely!

hardware_interface/include/hardware_interface/joint_mode_interface.h
bmagyar: good point

transmission_interface/include/transmission_interface/differential_transmission.h
bmagyar: I reworked this part, thanks for the suggestion. Now the tests don't need modifications anymore, they test the default value setting of not ignoring. We can add specific tests for the opposite later.

transmission_interface/include/transmission_interface/transmission.h
bmagyar: been aligned

transmission_interface/include/transmission_interface/simple_transmission.h
bmagyar: right!a

bmagyar:
> Also, a final comment: Why torque_sensor rather than torque?

For lack of a better alternative, this is how we try to show that this is a read-only value, not something ppl can use for commanding.

hardware_interface/include/hardware_interface/actuator_state_interface.h
bmagyar: done!

hardware_interface/include/hardware_interface/joint_mode_interface.h
bmagyar: It varies. At some point we should run a formatter on the entire codebase at once.

bmagyar:
I think I did all the requested changes, back to you ;)

hardware_interface/include/hardware_interface/joint_mode_interface.h
matthew-reynolds: Good call. Perhaps creating a `.clang-format` and applying it to the codebase should be added to #403

transmission_interface/include/transmission_interface/differential_transmission.h
matthew-reynolds: :+1:

transmission_interface/include/transmission_interface/transmission.h
matthew-reynolds: Got it

matthew-reynolds (approved):
Thanks for reworking the new `DifferentialTransmission` constructors, with that squared away looks good to me!

Looks good to merge, a couple tiny outstanding things:

 - [this](https://github.com/ros-controls/ros_control/pull/395#discussion_r361883779) comment about `actuatorToJointTorqueSensor()`'s function comment was not answered
 - [this](https://github.com/ros-controls/ros_control/pull/395#discussion_r361884538) and [this](https://github.com/ros-controls/ros_control/pull/395#discussion_r361883702) comment were marked resolved but the code wasn't updated. Not a big deal, especially if we plan on running a formatter soonish.

>>Also, a final comment: Why torque_sensor rather than torque?
>
> For lack of a better alternative, this is how we try to show that this is a read-only value, not something ppl can use for commanding.

Got it. Feels a bit inconsistent with `absolute_pos`, since they're both read-only sensors, but `absolute_position_sensor` is certainly quite wordy.

transmission_interface/include/transmission_interface/differential_transmission.h
bmagyar: my mistake, fixed!

transmission_interface/include/transmission_interface/differential_transmission.h
bmagyar: done

bmagyar:
The outstanding ones should be pushed, thanks for spotting them!

matthew-reynolds:
:+1: Looks great!

bmagyar:
Thanks a lot for the reviews!

matthew-reynolds:
Looking back at this again:

Did something change since https://github.com/pal-robotics-forks/ros_controllers/pull/41#issuecomment-72890364? From @adolfo-rt in that thread, Feb 2015:
>We've come to the conclusion that having controllers for handling joint mode switches is suboptimal

I'm wondering if there has been further discussion since then that prompted the `joint_mode_interface` to be merged in here, or if it should be pulled it back out.

bmagyar:
It wasn't an accident. After that conclusion, we also came to the realization that companies working with `ros_control` shouldn't keep internal forks solving this problem which forces them to ship their robots with custom, partially upstream-compatible ROS distributions.

Suboptimal is better than no solution at all.

matthew-reynolds:
Got it - Thanks for clarifying :)

PR #396, opened 2019-12-07
Small fixes

bmagyar:


matthew-reynolds:
This PR introduced the same issue resolved in ros-controls/ros_controllers#425.

@bmagyar @v-lopez Can you guys give a bit more info on why the change was made so that we can find a more robust solution? The commit message says the change was to resolve a build error in clang, but I'm able to build using clang 6.0 without these changes. Perhaps something else has changed in the 14 months since the original commit.

bmagyar:
This is all on me. I've been cherry picking old changes lately and it seems I overlooked this one. Would you mind reverting the specific parts that cause trouble?

matthew-reynolds:
All good - opened PR #398 to revert the changes

PR #397, opened 2019-12-12
[Windows][melodic-devel] `#undef ERROR` to avoid conflicts from Windows header.

seanyen:
This pull request is a minor edit for #391 to `#undef ERROR` to avoid conflicts from Windows header. The conflicts manifested when building [`gazebo_ros_control`](https://github.com/ros-simulation/gazebo_ros_pkgs) since two weeks ago.

### DIFF
--- a/hardware_interface/include/hardware_interface/robot_hw.h
+++ b/hardware_interface/include/hardware_interface/robot_hw.h
@@ -39,6 +39,10 @@
 #include <ros/console.h>
 #include <ros/node_handle.h>
 
+#ifdef ERROR
+#undef ERROR
+#endif
+
 namespace hardware_interface
 {
 
--- a/hardware_interface/include/hardware_interface/robot_hw.h
+++ b/hardware_interface/include/hardware_interface/robot_hw.h
@@ -39,7 +39,7 @@
 #include <ros/console.h>
 #include <ros/node_handle.h>
 
-// On Windows build, some namespaces has been taken by Windows header.
+// On Windows build, some namespaces have been taken by Windows header.
 // To avoid the conflicts, undefine it as needed.
 #ifdef ERROR
 #undef ERROR
--- a/hardware_interface/include/hardware_interface/robot_hw.h
+++ b/hardware_interface/include/hardware_interface/robot_hw.h
@@ -150,21 +150,21 @@ public:
 
   enum SwitchState
   {
-    SS_DONE,
-    SS_ONGOING,
-    SS_ERROR
+    DONE,
+    ONGOING,
+    ERROR
   };
 
   // Return (in realtime) the state of the last doSwitch()
   virtual SwitchState switchResult() const
   {
-    return SS_DONE;
+    return DONE;
   }
 
   // Return (in realtime) the state of the last doSwitch() for a given controller
   virtual SwitchState switchResult(const ControllerInfo& /*controller*/) const
   {
-    return SS_DONE;
+    return DONE;
   }
 
   /**
--- a/hardware_interface/include/hardware_interface/robot_hw.h
+++ b/hardware_interface/include/hardware_interface/robot_hw.h
@@ -150,21 +150,21 @@ public:
 
   enum SwitchState
   {
-    DONE,
-    ONGOING,
-    ERROR
+    SS_DONE,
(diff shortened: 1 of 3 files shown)


seanyen:
cc @bmagyar This is ready for review and merge. Hope the workaround is acceptable. Thanks!

bmagyar:
I think this sort of fix is more of a hack than a real fix.
Have you tried changing the enum to an enum class? I believe with those we can avoid dumping symbols into the global scope.

seanyen:
@bmagyar Thanks for the feedback. I tried to use `enum class`, but unfortunately it gave me the same result. However, when I change `ERROR` into `Error`, the `Pascal case` naming works. Would it be more proper?

mathias-luedtke:
> I think this sort of fix is more of a hack than a real fix.

I second that.

> I believe with those we can avoid dumping symbols into the global scope.

The current solution does not pollute the global scope (?).
Instead, some other other macro seems to pollute it.

> Have you tried changing the enum to an enum class?

This would still be an API break. On the other side, it is a recent addition, so better now than later.

mathias-luedtke:
> I tried to use enum class, but unfortunately it gave me the same result.

Instead, of changing it here, just fix that macro problem (which is the root cause), e.g. `undef` it.

hardware_interface/include/hardware_interface/robot_hw.h
```diff
@@ -150,21 +150,21 @@ class RobotHW : public InterfaceManager
 
   enum SwitchState
   {
-    DONE,
-    ONGOING,
-    ERROR
+    SS_DONE,
```
mathias-luedtke: This prefix really makes me cringe

mathias-luedtke:
BTW: The Google Style Guide recommends mixed case prefixed by `k`: https://google.github.io/styleguide/cppguide.html#Enumerator_Names

seanyen:
@ipa-mdl @bmagyar Thanks for the feedback. I changed the fix to `#undef ERROR` to avoid conflicts from Windows header.

mathias-luedtke:
@seanyen: Do you know which header actually injects the macro?

seanyen:
@ipa-mdl The short answer is `wingdi.h` where `ERROR` is defined.

The long version is that:
For `gazebo_ros_control`, many `.cpp` files include the Gazebo common headers (like, `gazebo/transport/Connection.hh`) and which subsequentially adds `boost/asio.hpp` and it internally includes `winsock2.h` and `windows.h` (see [here](https://www.boost.org/doc/libs/1_66_0/boost/asio/detail/socket_types.hpp)).

One another potential solution is to define `NOGDI` in `gazebo_ros_control` as a project-wide setting to tell `Windows` header to avoid this extra `wingdi.h` namespace pollution.

bmagyar:
@seanyen would you mind setting `NOGDI` in `gazebo_ros_control` instead? I'm very uneasy about adding protective measures to arbitrary things like this, if we can fight it off in cmake that'd be so much better.

mathias-luedtke:
>  would you mind setting NOGDI in gazebo_ros_control instead?

It might make sense to set `WIN32_LEAN_AND_MEAN` in general for all windows builds..

seanyen:
I proposed another fix in `gazebo_ros_control`: https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1023

seanyen:
> It might make sense to set `WIN32_LEAN_AND_MEAN` in general for all windows builds..

@ipa-mdl `WIN32_LEAN_AND_MEAN` is already enabled by default for Windows builds in [`catkin`](https://github.com/ros/catkin/blob/kinetic-devel/cmake/platform/windows.cmake#L97) today. However, like this case, there are still chances to see namespace conflicts, and we need to set `NOxxx` to further reduce the namespace pollution case by case.

mathias-luedtke:
@seanyen: Thanks for the clarification! I thought that `WIN32_LEAN_AND_MEAN` includes `NOGDI`.

bmagyar:
I take that this PR is no longer necessary, right?

seanyen:
> I take that this PR is no longer necessary, right?

Correct. I am closing this PR now.

PR #399, opened 2019-12-19
Fix compiler warnings

mvieth:
- Comment out unused parameters
- Make some integer literals unsigned to avoid comparison between signed and unsigned
- Remove unnecessary semicolons
- Make const void return type to void


bmagyar (approved):
Flying colours... Thanks for the dusting-off ;)

PR #401, opened 2019-12-20
Remove legacy gitignore workaround

matthew-reynolds:
Removes a workaround left from the rosbuild days and irrelevant in today's Catkin ecosystem.

See issue #101 and PR #108.


bmagyar (approved):
I in general never liked gitignore files anyways so the slimmer it is the better!

Issue #117, opened 2013-09-25
Make test dependencies optional


adolfo-rt:
Leverage the `CATKIN_ENABLE_TESTING` CMake variable (example) and the `<test_depend>` package.xml element.

bmagyar:
This has been long resolved.

PR #404, opened 2019-12-29
Update CMakeLists.txt and package.xml

matthew-reynolds:
Various fixes and formatting for `CMakeLists.txt` and `package.xml` files, including fixes suggested by `catkin_lint`.

---
#### Apply consistent format to CMakeLists.txt and package.xml files

The format is largely based on the existing files. The most important change is that this PR corrects the order of commands in the `CMakeLists.txt`.

I'm not at all attached to the format used in the CMakeLists.txt, happy to gut out the comments or make other stylistic changes as desired. I would mostly just like them to be consistent, as it makes it much easier to identify issues and to maintain.

----

#### Beyond stylistic changes, the following changes were made:

`combined_robot_hw`:
 - Remove unnecessary `boost` library dependency

`combined_robot_hw_tests`:
 - Remove unnecessary `boost` library dependency
 - Add missing `${catkin_EXPORTED_TARGETS}` dependency
 - Gut the `catkin_package()` and `install()` commands
 - Move all targets into the `if(CATKIN_ENABLE_TESTING)` block
 - Update package description
 - Add missing dependencies to `package.xml`

`controller_interface`:
 - Remove `rosunit` dependency (See note at bottom)
 - Remove unneeded pluginlib dependency

`controller_manager`:
 - Add missing `roscpp` dependency
 - Add missing `${catkin_EXPORTED_TARGETS}` dependency
 - Add missing `exec_depends` for python scripts

`controller_manager_msgs`:
 - Add missing `exec_depends` for python scripts

`controller_manager_tests`:
 - Remove unnecessary `boost` include_dir dependency
 - Add missing `hardware_interface` and `roscpp` dependencies
 - Add missing `${catkin_EXPORTED_TARGETS}` dependency
 - Gut the `catkin_package()` and `install()` commands
 - Move all targets into the `if(CATKIN_ENABLE_TESTING)` block
 - Update package description
 - Update dependencies in `package.xml`

`hardware_interface`:
 - Add missing `find_package(Boost REQUIRED)` for header-only library dependency `ptr_container`
 - Add missing `boost` depend in package.xml
 - Remove redundant `rosconsole` dependency (Already included via `roscpp`)
 - Remove unused `rostest` test_depend
 - Remove `rosunit` dependency (See note at bottom)

`joint_limits_interface`:
 - Remove `liburdfdom-dev` dependency (See https://github.com/ros-controls/ros_control/pull/404#discussion_r365019612)

`rqt_controller_manager`:
 - Install python scripts with `catkin_install_python()` rather than the default cmake `install()`
 - Update package description
 - Add missing dependencies

`transmission_interface`:
 - Add missing `find_package(Boost REQUIRED)` for header-only library dependency `lexical_cast`
 - Add missing `boost` depend in package.xml
 - Remove unused `Boost` dependencies in tests
 - Fix `find_package()` in tests, to not override previous `find_package(catkin [...])`. See https://github.com/ros/rosdistro/issues/3010#issuecomment-42116904

---

#### Notes:

I was unable to find up-to-date documentation on `rosunit` and the corresponding `catkin_add_gtest`/`catkin_add_gmock`. In particular, I don't know if it is necessary to `find_package(rosunit)` or `<test_depend>rosunit</test_depend>`. I believe the `find_package` is unecessary. I think the `test_depend` is required, but I'm not sure.

For header-only Boost libraries, I used `find_package(Boost REQUIRED)` and `include_directories(${Boost_INCLUDE_DIRS})`. I was not able to find any other examples of this use in large ROS packages, so I just wanted to highlight it here in case there's a better approach.

---

#### TODO:

Couple things I just want to look through before changing this from a draft to a normal PR:

 - ~Make sure `add_dependencies(${catkin_EXPORTED_TARGETS})` is used everywhere it's required~ Done in fa93166
 - ~Make sure `roscpp` is depended upon directly everywhere that it's required~ Done in 79b504c
 - ~Investigate the `urdf`/`liburdf-dev` situation, and see if things can be simplified~ Done in 7ca25ab

### DIFF
--- a/combined_robot_hw/CMakeLists.txt
+++ b/combined_robot_hw/CMakeLists.txt
@@ -1,33 +1,50 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(combined_robot_hw)
 
+# Load catkin and all dependencies required for this package
 find_package(catkin REQUIRED COMPONENTS
   hardware_interface
   pluginlib
   roscpp
 )
 
+# Declare a catkin package
+catkin_package(
+  INCLUDE_DIRS
+    include
+  LIBRARIES
+    ${PROJECT_NAME}
+  CATKIN_DEPENDS
+    hardware_interface
+    pluginlib
+    roscpp
+)
+
+
+###########
+## Build ##
+###########
+
+# Specify header include paths
 include_directories(include ${catkin_INCLUDE_DIRS})
 
-catkin_package(
-  INCLUDE_DIRS include
-  LIBRARIES ${PROJECT_NAME}
-  CATKIN_DEPENDS hardware_interface pluginlib roscpp
-)
-
-add_library(${PROJECT_NAME}
-  src/combined_robot_hw.cpp
-)
+add_library(${PROJECT_NAME} src/combined_robot_hw.cpp)
 add_dependencies(${PROJECT_NAME} ${catkin_EXPORTED_TARGETS})
 target_link_libraries(${PROJECT_NAME} ${catkin_LIBRARIES} ${Boost_LIBRARIES})
 
 
-# Install
-install(DIRECTORY include/${PROJECT_NAME}/
-  DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION})
+#############
+## Install ##
+#############
 
-# Install library
+# Install headers
+install(DIRECTORY include/${PROJECT_NAME}/
+  DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}
+)
+
+# Install targets
 install(TARGETS ${PROJECT_NAME}
   ARCHIVE DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
   LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
-  RUNTIME DESTINATION ${CATKIN_GLOBAL_BIN_DESTINATION})
+  RUNTIME DESTINATION ${CATKIN_GLOBAL_BIN_DESTINATION}
+)
--- a/combined_robot_hw/CMakeLists.txt
+++ b/combined_robot_hw/CMakeLists.txt
@@ -30,7 +30,7 @@ include_directories(include ${catkin_INCLUDE_DIRS})
 
 add_library(${PROJECT_NAME} src/combined_robot_hw.cpp)
 add_dependencies(${PROJECT_NAME} ${catkin_EXPORTED_TARGETS})
-target_link_libraries(${PROJECT_NAME} ${catkin_LIBRARIES} ${Boost_LIBRARIES})
+target_link_libraries(${PROJECT_NAME} ${catkin_LIBRARIES})
 
 
 #############
--- a/combined_robot_hw_tests/CMakeLists.txt
+++ b/combined_robot_hw_tests/CMakeLists.txt
@@ -32,7 +32,6 @@ if(CATKIN_ENABLE_TESTING)
     src/my_robot_hw_3.cpp
     src/my_robot_hw_4.cpp
   )
-  add_dependencies(${PROJECT_NAME} ${catkin_EXPORTED_TARGETS})
   target_link_libraries(${PROJECT_NAME} ${catkin_LIBRARIES})
 
   add_executable(combined_robot_hw_dummy_app EXCLUDE_FROM_ALL src/dummy_app.cpp)
@@ -48,7 +47,7 @@ if(CATKIN_ENABLE_TESTING)
     test/cm_test.test
     test/cm_test.cpp
   )
-  add_dependencies(combined_robot_hw_cm_test combined_robot_hw_dummy_app)
+  add_dependencies(combined_robot_hw_cm_test combined_robot_hw_dummy_app ${catkin_EXPORTED_TARGETS})
   target_link_libraries(combined_robot_hw_cm_test ${PROJECT_NAME} ${catkin_LIBRARIES})
 endif()
 
--- a/combined_robot_hw_tests/CMakeLists.txt
+++ b/combined_robot_hw_tests/CMakeLists.txt
@@ -11,37 +11,7 @@ find_package(catkin REQUIRED COMPONENTS
 )
 
 # Declare a catkin package
-catkin_package(
-  INCLUDE_DIRS
-    include
-  LIBRARIES
-    ${PROJECT_NAME}
-  CATKIN_DEPENDS
-    combined_robot_hw
-    hardware_interface
-    roscpp
-    # Note: controller_manager and controller_manager_tests are only used in tests
-)
-
-
-###########
-## Build ##
-###########
-
-# Specify header include paths
-include_directories(include ${catkin_INCLUDE_DIRS})
-
-add_library(${PROJECT_NAME}
-  src/my_robot_hw_1.cpp
-  src/my_robot_hw_2.cpp
-  src/my_robot_hw_3.cpp
-  src/my_robot_hw_4.cpp
-)
-add_dependencies(${PROJECT_NAME} ${catkin_EXPORTED_TARGETS})
-target_link_libraries(${PROJECT_NAME} ${catkin_LIBRARIES})
-
-add_executable(combined_robot_hw_dummy_app src/dummy_app.cpp)
-target_link_libraries(combined_robot_hw_dummy_app ${PROJECT_NAME} ${catkin_LIBRARIES})
+catkin_package()
 
 
 #############
@@ -50,6 +20,21 @@ target_link_libraries(combined_robot_hw_dummy_app ${PROJECT_NAME} ${catkin_LIBRA
 
 if(CATKIN_ENABLE_TESTING)
   find_package(rostest REQUIRED)
+
+  # Specify header include paths
+  include_directories(include ${catkin_INCLUDE_DIRS})
+
+  add_library(${PROJECT_NAME} EXCLUDE_FROM_ALL
+    src/my_robot_hw_1.cpp
+    src/my_robot_hw_2.cpp
+    src/my_robot_hw_3.cpp
+    src/my_robot_hw_4.cpp
+  )
+  add_dependencies(${PROJECT_NAME} ${catkin_EXPORTED_TARGETS})
+  target_link_libraries(${PROJECT_NAME} ${catkin_LIBRARIES})
+
+  add_executable(combined_robot_hw_dummy_app EXCLUDE_FROM_ALL src/dummy_app.cpp)
+  target_link_libraries(combined_robot_hw_dummy_app ${PROJECT_NAME} ${catkin_LIBRARIES})
 
   add_rostest_gtest(combined_robot_hw_test
     test/combined_robot_hw_test.test
@@ -63,25 +48,3 @@ if(CATKIN_ENABLE_TESTING)
   )
   target_link_libraries(combined_robot_hw_cm_test ${PROJECT_NAME} ${catkin_LIBRARIES})
 endif()
-
-
-#############
-## Install ##
-#############
-
-# Install headers
-install(DIRECTORY include/${PROJECT_NAME}/
-  DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}
-)
-
-# Install targets
-install(TARGETS ${PROJECT_NAME} combined_robot_hw_dummy_app
-  ARCHIVE DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
-  LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
-  RUNTIME DESTINATION ${CATKIN_GLOBAL_BIN_DESTINATION}
-)
-
-# Install plugins
-install(FILES test_robot_hw_plugin.xml
-  DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}
-)
--- a/combined_robot_hw_tests/CMakeLists.txt
+++ b/combined_robot_hw_tests/CMakeLists.txt
@@ -1,6 +1,7 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(combined_robot_hw_tests)
 
+# Load catkin and all dependencies required for this package
 find_package(catkin REQUIRED COMPONENTS
   combined_robot_hw
   controller_manager
@@ -9,13 +10,26 @@ find_package(catkin REQUIRED COMPONENTS
   roscpp
 )
 
-include_directories(include ${catkin_INCLUDE_DIRS})
-
+# Declare a catkin package
 catkin_package(
-  INCLUDE_DIRS include
-  LIBRARIES ${PROJECT_NAME}
-  CATKIN_DEPENDS combined_robot_hw hardware_interface roscpp
+  INCLUDE_DIRS
+    include
+  LIBRARIES
+    ${PROJECT_NAME}
+  CATKIN_DEPENDS
+    combined_robot_hw
+    hardware_interface
+    roscpp
+    # Note: controller_manager and controller_manager_tests are only used in tests
 )
+
+
+###########
+## Build ##
+###########
+
+# Specify header include paths
+include_directories(include ${catkin_INCLUDE_DIRS})
 
 add_library(${PROJECT_NAME}
   src/my_robot_hw_1.cpp
@@ -29,9 +43,14 @@ target_link_libraries(${PROJECT_NAME} ${catkin_LIBRARIES} ${Boost_LIBRARIES})
 add_executable(combined_robot_hw_dummy_app src/dummy_app.cpp)
 target_link_libraries(combined_robot_hw_dummy_app ${PROJECT_NAME} ${catkin_LIBRARIES})
 
-if(CATKIN_ENABLE_TESTING)
 
+#############
+## Testing ##
+#############
+
+if(CATKIN_ENABLE_TESTING)
   find_package(rostest REQUIRED)
+
   add_rostest_gtest(combined_robot_hw_test
     test/combined_robot_hw_test.test
     test/combined_robot_hw_test.cpp
@@ -43,18 +62,26 @@ if(CATKIN_ENABLE_TESTING)
     test/cm_test.cpp
   )
   target_link_libraries(combined_robot_hw_cm_test ${PROJECT_NAME} ${catkin_LIBRARIES})
-
 endif()
 
-# Install
-install(DIRECTORY include/${PROJECT_NAME}/
-  DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION})
 
-# Install library
-install(TARGETS ${PROJECT_NAME}
+#############
+## Install ##
+#############
+
+# Install headers
+install(DIRECTORY include/${PROJECT_NAME}/
+  DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}
+)
+
+# Install targets
+install(TARGETS ${PROJECT_NAME} combined_robot_hw_dummy_app
   ARCHIVE DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
   LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
-  RUNTIME DESTINATION ${CATKIN_GLOBAL_BIN_DESTINATION})
+  RUNTIME DESTINATION ${CATKIN_GLOBAL_BIN_DESTINATION}
+)
 
+# Install plugins
 install(FILES test_robot_hw_plugin.xml
-  DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION})
+  DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}
+)
--- a/combined_robot_hw_tests/CMakeLists.txt
+++ b/combined_robot_hw_tests/CMakeLists.txt
@@ -38,7 +38,7 @@ add_library(${PROJECT_NAME}
   src/my_robot_hw_4.cpp
 )
 add_dependencies(${PROJECT_NAME} ${catkin_EXPORTED_TARGETS})
-target_link_libraries(${PROJECT_NAME} ${catkin_LIBRARIES} ${Boost_LIBRARIES})
+target_link_libraries(${PROJECT_NAME} ${catkin_LIBRARIES})
 
 add_executable(combined_robot_hw_dummy_app src/dummy_app.cpp)
 target_link_libraries(combined_robot_hw_dummy_app ${PROJECT_NAME} ${catkin_LIBRARIES})
--- a/combined_robot_hw_tests/CMakeLists.txt
+++ b/combined_robot_hw_tests/CMakeLists.txt
@@ -49,3 +49,13 @@ if(CATKIN_ENABLE_TESTING)
   add_dependencies(combined_robot_hw_cm_test combined_robot_hw_dummy_app)
   target_link_libraries(combined_robot_hw_cm_test ${PROJECT_NAME} ${catkin_LIBRARIES})
 endif()
+
+
+#############
+## Install ##
+#############
+
+# Install plugins
+install(FILES test_robot_hw_plugin.xml
+  DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}
+)
--- a/combined_robot_hw_tests/CMakeLists.txt
+++ b/combined_robot_hw_tests/CMakeLists.txt
@@ -5,8 +5,10 @@ project(combined_robot_hw_tests)
 find_package(catkin REQUIRED COMPONENTS
   combined_robot_hw
   controller_manager
+  controller_manager_msgs
   controller_manager_tests
   hardware_interface
+  pluginlib
   roscpp
 )
 
--- a/combined_robot_hw_tests/CMakeLists.txt
+++ b/combined_robot_hw_tests/CMakeLists.txt
@@ -26,7 +26,7 @@ if(CATKIN_ENABLE_TESTING)
   # Specify header include paths
   include_directories(include ${catkin_INCLUDE_DIRS})
 
-  add_library(${PROJECT_NAME} EXCLUDE_FROM_ALL
+  add_library(${PROJECT_NAME}
     src/my_robot_hw_1.cpp
     src/my_robot_hw_2.cpp
     src/my_robot_hw_3.cpp
@@ -57,7 +57,20 @@ endif()
 ## Install ##
 #############
 
-# Install plugins
-install(FILES test_robot_hw_plugin.xml
-  DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}
-)
+# NOTE: Libraries and plugins required for tests are installed since CI
+# runs tests out of the install space rather than the devel space
+
+if(CATKIN_ENABLE_TESTING)
+
+  # Install targets
+  install(TARGETS ${PROJECT_NAME}
+    ARCHIVE DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
+    LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
+    RUNTIME DESTINATION ${CATKIN_GLOBAL_BIN_DESTINATION}
+  )
+
+  # Install plugins
+  install(FILES test_robot_hw_plugin.xml
+    DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}
+  )
+endif()
--- a/combined_robot_hw_tests/CMakeLists.txt
+++ b/combined_robot_hw_tests/CMakeLists.txt
@@ -46,5 +46,6 @@ if(CATKIN_ENABLE_TESTING)
     test/cm_test.test
     test/cm_test.cpp
   )
+  add_dependencies(combined_robot_hw_cm_test combined_robot_hw_dummy_app)
   target_link_libraries(combined_robot_hw_cm_test ${PROJECT_NAME} ${catkin_LIBRARIES})
 endif()
--- a/controller_manager/CMakeLists.txt
+++ b/controller_manager/CMakeLists.txt
@@ -2,65 +2,92 @@ cmake_minimum_required(VERSION 2.8.3)
 project(controller_manager)
 
 # Load catkin and all dependencies required for this package
-find_package(catkin REQUIRED COMPONENTS controller_interface controller_manager_msgs hardware_interface pluginlib)
+find_package(catkin REQUIRED COMPONENTS
+  controller_interface
+  controller_manager_msgs
+  hardware_interface
+  pluginlib
+)
 
+# Include Boost.Thread
 find_package(Boost REQUIRED COMPONENTS thread)
 
-include_directories(include ${Boost_INCLUDE_DIR} ${catkin_INCLUDE_DIRS})
+catkin_python_setup()
 
-# Declare catkin package
+# Declare a catkin package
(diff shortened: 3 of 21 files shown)


combined_robot_hw/CMakeLists.txt
```diff
@@ -1,33 +1,50 @@
...
-add_library(${PROJECT_NAME}
-  src/combined_robot_hw.cpp
-)
+
+###########
+## Build ##
+###########
+
+# Specify header include paths
+include_directories(include ${catkin_INCLUDE_DIRS})
```
bmagyar, line 29: what do you think about using `target_include_directories` instead of this global option?

combined_robot_hw_tests/CMakeLists.txt
```diff
@@ -43,18 +62,26 @@ if(CATKIN_ENABLE_TESTING)
...
+# Install headers
 install(DIRECTORY include/${PROJECT_NAME}/
-  DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION})
+  DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}
+)
 
-# Install library
-install(TARGETS ${PROJECT_NAME}
+# Install targets
+install(TARGETS ${PROJECT_NAME} combined_robot_hw_dummy_app
```
bmagyar: the dummy app is only for tests, dont think we need to install

combined_robot_hw_tests/CMakeLists.txt
```diff
@@ -9,29 +10,47 @@ find_package(catkin REQUIRED COMPONENTS
...
-if(CATKIN_ENABLE_TESTING)
 
+#############
+## Testing ##
+#############
+
+if(CATKIN_ENABLE_TESTING)
   find_package(rostest REQUIRED)
+
   add_rostest_gtest(combined_robot_hw_test
```
bmagyar, line 40: is there a way to `add_dependencies(.... combined_robot_hw_dummy_app)`?

controller_manager_tests/CMakeLists.txt
```diff
@@ -28,30 +42,45 @@ target_link_libraries(${PROJECT_NAME} ${catkin_LIBRARIES})
...
+# Install headers
 install(DIRECTORY include/${PROJECT_NAME}/
-  DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION})
+  DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}
+)
 
-# Install library
-install(TARGETS ${PROJECT_NAME}
+# Install targets
+install(TARGETS ${PROJECT_NAME} dummy_app
```
bmagyar: again, not sure it's good to install this

hardware_interface/package.xml
```diff
@@ -16,7 +17,8 @@
   <author>Adolfo Rodriguez Tsouroukdissian</author>
 
   <buildtool_depend>catkin</buildtool_depend>
+
   <depend>roscpp</depend>
-  <test_depend>rostest</test_depend>
+
   <test_depend>rosunit</test_depend>
```
bmagyar: I'm wondering what rosunit brings us at this stage... gmock is integrated into catkin

joint_limits_interface/package.xml
```diff
@@ -2,22 +2,26 @@
...
+
   <author>Adolfo Rodriguez Tsouroukdissian</author>
 
+
   <buildtool_depend>catkin</buildtool_depend>
 
-  <depend>roscpp</depend>
   <depend>hardware_interface</depend>
-  <depend>urdf</depend>
   <depend>liburdfdom-dev</depend>
```
bmagyar: Could you please have a look at which packages can be reduced to a build_depend? I think everything with a `-dev` is a good candidate

transmission_interface/package.xml
```diff
@@ -17,17 +17,18 @@
 
   <buildtool_depend>catkin</buildtool_depend>
 
-  <depend>tinyxml</depend>
-  <depend>roscpp</depend>
   <depend>pluginlib</depend>
-  <build_depend>hardware_interface</build_depend>
+  <depend>roscpp</depend>
+  <depend>tinyxml</depend>
```
bmagyar, line 21: Is this a header-only tinyxml version by any chance?

transmission_interface/package.xml
```diff
@@ -17,17 +17,18 @@
...
+
   <build_depend>cmake_modules</build_depend>
+  <build_depend>hardware_interface</build_depend>
 
-  <test_depend>rosunit</test_depend>
   <test_depend>resource_retriever</test_depend>
+  <test_depend>rosunit</test_depend>
 
   <export>
     <rosdoc config="rosdoc.yaml"/>
```
bmagyar, line 35: I don't think rosdoc is being used anymore

bmagyar (changes requested):

combined_robot_hw/CMakeLists.txt
matthew-reynolds: I am not a cmake & catkin expert, but my understanding is that within the framework of catkin, most of the benefits of `target_include_directories` are lost. Here is my understanding of the situation - Please correct me if I'm wrong:

The main "disadvantage" of `include_directories` is that it offers less granularity, affecting all targets in the current file. More importantly, it also affects all targets in files included via `add_subdirectory`. Since we don't have cmake subdirectories (thanks to catkin), we do not need to worry about this second point. The first point is valid, but the lack of control between targets doesn't seem to be a big deal IMO.

`target_include_directories` offers the ability to control what headers are exported to dependent targets. However, the `catkin_package(INCLUDE_DIRS)` here and `${catkin_INCLUDE_DIRS}` in dependent packages handles that for us.

In my (very non-expert) view, since all the benefits are already handled by catkin, `target_include_directories` would just clutter the `CMakeLists` without any benefits.

combined_robot_hw_tests/CMakeLists.txt
matthew-reynolds: Got it. In that case, should I also remove the install for `${PROJECT_NAME}`? It seems to me we should either install all targets or none, and none seems to make more sense. Should I also remove the exported headers and targets from the `catkin_package()` call?

Same for headers and plugins - Do we need to install anything at all?

These test-only packages still feel quite odd to me. Why are they separate packages at all, and more importantly, why are they released and dependencies of `ros_control`? There are only a couple of packages with 'test' in their name in [rosdistro](https://github.com/ros/rosdistro/blob/master/melodic/distribution.yaml), and as far as I can tell we have almost the only packages that are just rostest & rosunit tests.

combined_robot_hw/CMakeLists.txt
bmagyar: Point taken. We are on the same page about this and indeed the even though the global approach is a little sloppier, it saves boilerplate that wouldn't change much anyways.

combined_robot_hw_tests/CMakeLists.txt
bmagyar: I guess it depends on how one views it, should these be installable to then run as standalone tests or not? I think that since most of the framework has strong API dependencies internally, these would need to be recompiled anyways and if needed, they can be locally compiled. 

The original reasoning for these test-only packages was to have an all-round integration suite, these packages export dummy robothw components, controllers, etc. 

I'd opt for the choice of not installing or exporting headers from these packages.

I'm also in favor of removing them from the ros_control metapackage however I am not entirely sure we can get rid of them in the rosdistro file, the release scripts will probably want to update them all and bloom will look for them too. Let's try gutting them first.

hardware_interface/package.xml
bmagyar: I took a look. I think it is safe to remove, https://github.com/ros/ros/tree/kinetic-devel/tools/rosunit

hardware_interface/CMakeLists.txt
```diff
@@ -1,54 +1,76 @@
...
 
-find_package(catkin REQUIRED COMPONENTS roscpp)
+# Load catkin and all dependencies required for this package
+find_package(catkin REQUIRED COMPONENTS
+  roscpp
+)
 
-# Declare catkin package
+# Include header-only Boost libraries
+find_package(Boost REQUIRED)
```
bmagyar, line 10: off of the top of my head, this could read something similar
```suggestion
find_package(Boost REQUIRED COMPONENTS PointerContainer)
```

transmission_interface/CMakeLists.txt
```diff
@@ -9,117 +9,133 @@ find_package(catkin REQUIRED COMPONENTS
   roscpp
 )
 
+# Include header-only Boost libraries
+find_package(Boost REQUIRED)
```
bmagyar, line 13: Another guess
```suggestion
find_package(Boost REQUIRED COMPONENTS LexicalCast)
```

bmagyar:
@matthew-reynolds is there anything else you'd add apart from the outstanding review points? (The PR is still in draft mode)

matthew-reynolds:
Been quite busy the last 2 weeks, my apologies. Other than the outstanding review points, I just want to double check those last couple points in the initial PR message (that we're not missing and `roscpp` or `${catkin_EXPORTED_TARGETS}` dependencies). I'll try to get through this in the next couple days :)

hardware_interface/CMakeLists.txt
matthew-reynolds: According to the first comment on [this SO answer](https://stackoverflow.com/a/6646518), you don't need to `find_package` any components which are header-only (Such as pointer container and lexical cast)

transmission_interface/CMakeLists.txt
matthew-reynolds: See above

transmission_interface/package.xml
matthew-reynolds: What do you mean? The autogenerated documentation is still being autogenerated: http://docs.ros.org/api/ and http://docs.ros.org/melodic/api/transmission_interface/html/c++/.

Are you maybe thinking of the rosdoc -> rosdoc_lite transition between Groovy and Indigo?

combined_robot_hw_tests/CMakeLists.txt
matthew-reynolds: Good catch, I'll add that to the tests which use it

hardware_interface/package.xml
matthew-reynolds: [This doc](https://github.com/ros/catkin/blob/kinetic-devel/doc/howto/format2/gtest_configuration.rst) seems to indicate that it's still required, but that doc also looks horribly out of date (It looks to just be cloned over from package format 1, and untouched since 2014).

Ignoring that out-of-date doc, I think you're right. Thanks for looking into it.

I also did a quick sanity check against some other packages in the `ros` GitHub org. There aren't any `catkin_add_gmock`, but they have a lot of `catkin_add_gtest` (which is defined in [the same file](https://github.com/ros/catkin/blob/kinetic-devel/cmake/test/gtest.cmake)).

It's quite inconsistent (Likely because of legacy, I didn't dig through each package), but it appears the dependency is not needed nor standard.

I'm comfortable removing it if you are as well. Perhaps I should open up a PR to update that doc I linked above...

| Package | `catkin_add_gtest` | `find_package(rosunit)` | `<depend>rosunit</depend>` (package.xml only, excludes manifest.xml) |
| --- | :---: | :---: | :---: |
| actionlib |  ✔️ | ✔️ | ❌ |
| allocators |  ✔️ | ❌ | ❌ |
| angles |  ✔️ | ❌ | ✔️ |
| class_loader |  ✔️ | ❌ | ❌ |
| collada_urdf |  ✔️ | ❌ | ❌ |
| filters |  ✔️ | ❌ | ❌ |
| lockfree |  ✔️ | ❌ | ❌ |
| message_filters |  ✔️ | ✔️ | ✔️ |
| plugin_lib |  ✔️ | ❌ | ❌ |
| resource_retriever |  ✔️ | ❌ | ❌ |
| rosatomic |  ✔️ | ❌ | ❌ |
| rosbag_image_compressor |  ✔️ | ❌ | ❌ |
| rosbag_storage |  ✔️ | ❌ | ❌ |
| rosconsole |  ✔️ | ✔️ | ✔️ |
| roslaunch | ❌ (Uses `catkin_add_nosetests`) | ❌ | ✔️ |
| roslib |  ✔️ | ❌ | ❌ |
| roslz4 |  ✔️ | ❌ | ✔️ |
| rospack |  ✔️ | ❌ | ❌ |
| rosrt |  ✔️ | ✔️ | ✔️ |
| rostest |  ✔️ | ✔️ | ✔️ |
| rostime |  ✔️ | ❌ | ❌ |
| sensor_msgs |  ✔️ | ❌ | ✔️ |
| test_nodelet |  ✔️ | ❌ | ❌ |
| test_rosbag |  ✔️ | ✔️ | ✔️ |
| test_rosbag_storage |  ✔️ | ❌ | ❌ |
| test_roscpp |  ✔️ | ✔️ | ✔️ |
| test_rosservice | ❌ (Uses `catkin_add_nosetests`) | ✔️ | ✔️ |
| test_rospy | ❌ (Uses `catkin_add_nosetests`) | ✔️ | ❌ |
| test_tf2 |  ✔️ | ❌ | ✔️ |
| tf |  ✔️ | ❌ | ✔️ |
| tf_conversions |  ✔️ | ❌ | ❌ |
| tf2 |  ✔️ | ❌ | ❌ |
| tf2_bullet |  ✔️ | ❌ | ❌ |
| tf2_eigen |  ✔️ | ❌ | ❌ |
| tf2_geometry_msgs |  ✔️ | ❌ | ❌ |
| tf2_py |  ✔️ | ❌ | ❌ |
| topic_tools |  ✔️ | ✔️ | ✔️ |
| urdf |  ✔️ | ❌ | ❌ |
| xmlrpcpp |  ✔️ | ❌ | ❌ |

combined_robot_hw_tests/CMakeLists.txt
matthew-reynolds: > The original reasoning for these test-only packages was to have an all-round integration suite, these packages export dummy robothw components, controllers, etc.

Do you mean so that people who write plugins and extensions have some tools to test their code? That would be cool, I saw a couple packages like that when I looked through rosdistro. I agree that they're not really offering that as they are though.

> I'm also in favor of removing them from the ros_control metapackage however I am not entirely sure we can get rid of them in the rosdistro file

Agreed. Un-releasing packages doesn't seem possible or desirable, even if they're mostly useless to the outside world. Should we remove them from the metapackage in Melodic (And maybe Kinetic), or only in Noetic+ to stay consistent with the rosdistro file? I would lean towards removing it today.

> I'd opt for the choice of not installing or exporting headers from these packages

Sounds good. I will also remove the plugin export.

---

A final note: Shall I open an issue to track this so we can keep improving the situation? Whether that means updating the metapackage or pulling the tests into their corresponding packages (leaving empty packages for now, to be removed in Noetic) or whatever else we decide.

Fixing the CMakeLists.txt is certainly in scope for this PR, but playing with it any further feels like it deserves its own PR.

hardware_interface/CMakeLists.txt
bmagyar: That makes a lot of sense, I'll keep this in mind. Thanks!

transmission_interface/package.xml
bmagyar: You are right, for some reason I was under the impression these additional yaml files were ignored since Indigo.

combined_robot_hw_tests/CMakeLists.txt
matthew-reynolds: Looking at `test_roscpp` and the other `test_XXX` packages in `ros_comm` as an example for these "test-packages," it seems we are on the right track by gutting the CMakeLists.txt of any installs and exports, and eventually halting distribution of the packages altogether.

Following the example from those packages, we should also be using `EXCLUDE_FROM_ALL` on our `add_executable` and `add_library` targets. We should also move all of our `include_directory` and `add_executable`/`add_library` stuff into the `if(CATKIN_ENABLE_TESTING)`, leaving just the stuff like `catkin_package` outside the if.

I will follow the example of these packages and update our `_tests` packages accordingly.

matthew-reynolds: Fixed in 6f4648de

combined_robot_hw_tests/CMakeLists.txt
matthew-reynolds: Added in 41f6cff8

controller_manager_tests/CMakeLists.txt
matthew-reynolds: Fixed in 6f4648d

hardware_interface/package.xml
matthew-reynolds: Fixed in 37d1288e

matthew-reynolds:
While finding missing roscpp & rospy deps, I found a couple additional missing package dependencies (some msgs, pluginlib in some tests, etc). Will do another thorough pass to fix all of those.

Still need to double check we're not missing any `${catkin_EXPORTED_TARGETS}` cmake dependencies, and go over the urdf & tinyxml situation again.

combined_robot_hw_tests/CMakeLists.txt
matthew-reynolds: Looks like tests fail if we don't install plugins. Will investigate.

matthew-reynolds: I'm able to reproduce the issue reliably using `industrial_ci` locally, but I haven't been able to reproduce it using `catkin run_tests` directly, even doing my best to follow the steps `industrial_ci` takes.

Regardless, I suspect the issue may be related to `catkin_tools` [incorrectly sourcing the devel space](https://github.com/catkin/catkin_tools/issues/397) when running tests. I haven't fully investigated, but it seems the easiest solution is probably to just install these plugins. Unless you've got a better idea, I'm going to revert back to installing just the plugins.

transmission_interface/package.xml
matthew-reynolds: I'm not aware of any header-only TinyXML versions. This is just the system-provided TinyXML w/ the ROS `cmake_modules`.

joint_limits_interface/package.xml
matthew-reynolds: `liburdfdom-dev` is a dependency of `urdf`, and our use of it explicitly is a leftover from [a 2014 packaging issue](https://github.com/ros/rosdistro/issues/4633). That issue is long resolved, and Indigo is EOL, so I think we're good to drop it (7ca25abf).

I will go through the other dependencies and see what can be reduced though.

matthew-reynolds: Went through and update all the dependencies in b3654b3b

combined_robot_hw_tests/CMakeLists.txt
matthew-reynolds: Re-installed in 2864284b

hardware_interface/package.xml
matthew-reynolds: `catkin_lint` now complains that we're missing a `test_depend` on `rosunit`.

Perhaps we should get some outside help to clarify best practices here.

combined_robot_hw_tests/CMakeLists.txt
matthew-reynolds: Test still fails because now the plugin can't find its corresponding library, since it's not installed.

It seems we're stuck installing all the bits related to plugins, since tests are run out of the install space rather than the devel space.

matthew-reynolds: Update: I can reproduce the issue if I run the tests out of the install space.

This behaviour is consistent with `industrial_ci`. Essentially:

```
catkin config --install
catkin build
catkin run_tests
```

As I understand the issue, since `industrial_ci` uses the install space and `catkin_tools` doesn't switch to the devel space when running tests (as discussed in the issue linked above), the only solution is to install the plugins and corresponding libraries required by the tests.

bmagyar: Please open a follow-up issue for this and let's continue the discussion there.

combined_robot_hw_tests/CMakeLists.txt
bmagyar: Neat, thanks!

joint_limits_interface/package.xml
bmagyar: Thank you!

transmission_interface/package.xml
bmagyar: Right. I've dug in a bit and found that the `tinyxml` rosdep symbol resolves to `libtinyxml-dev` which also pulls in the library. Not ideal, as one would not want to carry the `-dev` package with them as a `run_depend` but it seems we don't have a choice.

hardware_interface/package.xml
```diff
@@ -16,7 +17,7 @@
   <author>Adolfo Rodriguez Tsouroukdissian</author>
 
   <buildtool_depend>catkin</buildtool_depend>
+
+  <depend>boost</depend>
```
bmagyar: perhaps only build dep here?

transmission_interface/package.xml
```diff
@@ -17,17 +17,18 @@
 
   <buildtool_depend>catkin</buildtool_depend>
 
-  <depend>tinyxml</depend>
-  <depend>roscpp</depend>
+  <depend>boost</depend>
```
bmagyar: build-depend?

hardware_interface/package.xml
bmagyar: Let's make a separate issue for this, this PR has reached a state where I'm eager to merge already ;)

hardware_interface/package.xml
matthew-reynolds: It's used in `hardware_inteface/internal/interface_manager.h`, which is included by `hardware_interface/robot_hw.h`

transmission_interface/package.xml
matthew-reynolds: It's used in `transmission_interface/transmission_interface_loader.h`

hardware_interface/package.xml
matthew-reynolds: Sounds good. Would you rather leave it as is, with no `rosunit` dependency and `catkin_lint` complaints, or would you prefer I re-add the dep for now?

combined_robot_hw_tests/CMakeLists.txt
matthew-reynolds: Will do.

hardware_interface/package.xml
matthew-reynolds: Follow up in #408.

I'm pretty sure we're good without `rosunit` and that `catkin_lint` is just out of date (The rule for `catkin_add_gtest` hasn't been updated in 5 years), so I'm leaning towards leaving the code as is for this PR since I believe we likely have the correct configuration.

combined_robot_hw_tests/CMakeLists.txt
matthew-reynolds: Issues #409 and #410 opened.

matthew-reynolds:
Follow-up issues all opened. Ready for final review of this PR.

Would you like me to reference #410 in the comment I left in the two _tests CMakeLists.txt?

BTW, came across these while working on this PR: #117 and #157 both look like old issues that have long been resolved. I think they should be closed.

transmission_interface/package.xml
bmagyar: Yes, but a header-only dependency doesn't need to be propagated as a runtime dependency. I'm merely proposing to
```suggestion
  <build_depend>boost</build_depend>
```

hardware_interface/package.xml
bmagyar: Just to clarify my suggestion :) 
```suggestion
  <build_depend>boost</build_depend>
```

hardware_interface/package.xml
bmagyar: If the CI passes, we should be ok.

bmagyar:
Thanks very much for sticking through this long review.
No worries, we are fine without that comment, the issue is open here.
I've closed those two outdated ones, thanks for bringing it to my attention!

transmission_interface/package.xml
matthew-reynolds: Whoops, yes you are correct.

matthew-reynolds: Actually, in this case we need a build_export_depend, since the dependency is used in a public header of this package. I'll give another pass through all the dependencies.

`build_depend` on packages needed to compile (headers & shared libraries)
`build_export_depend` on packages whose headers you transitively include in your public headers
`exec_depend` on packages needed to link (shared libraries)

bmagyar: I think the `internal` header above could use just build and build_export_depend for this case

matthew-reynolds: Agreed. I found a handful of other mistakes in my last pass at updating dependencies though. In particular, a lot of header-only deps should be updated to build & build_export depend. Currently working on it.

bmagyar:
Apologies for the abrupt "merge branch..." solution, I dared to resolve the conflict using github's interface, please feel free to squash that with any of your commits.

controller_manager/CMakeLists.txt
```diff
@@ -2,65 +2,96 @@ cmake_minimum_required(VERSION 2.8.3)
...
+    Boost
 )
 
+
+###########
+## Build ##
+###########
+
+# Specify header include paths
+include_directories(include ${catkin_INCLUDE_DIRS} ${Boost_INCLUDE_DIR})
```
gavanderhoorn: From the docs on `FindBoost.cmake` ([these](https://cmake.org/cmake/help/v3.0/module/FindBoost.html)) I get the impression `Boost_INCLUDE_DIRS` (note the `S`) should be used here. See the examples near the bottom of the page.

hardware_interface/CMakeLists.txt
```diff
@@ -1,54 +1,78 @@
...
 catkin_package(
-  CATKIN_DEPENDS roscpp
-  INCLUDE_DIRS include
-  )
+  INCLUDE_DIRS
+    include
+  CATKIN_DEPENDS
+    roscpp
+  DEPENDS
+    Boost # TODO: Should header-only dependencies be DEPENDS or INCLUDE_DIRS?
```
gavanderhoorn: The include paths will be automatically exported by this `DEPENDS Boost`.

Additionally, `INCLUDE_DIRS` is for (from [here](http://docs.ros.org/melodic/api/catkin/html/dev_guide/generated_cmake_api.html#catkin-package)):

```
CMAKE_CURRENT_SOURCE_DIR-relative paths to C/C++ includes
```

The include directories of `Boost` do not fall into that category.

transmission_interface/package.xml
gavanderhoorn: Could consider contributing a `rosdep` rule to split the runtime from the devel package?

controller_manager/CMakeLists.txt
matthew-reynolds: Good catch!

hardware_interface/CMakeLists.txt
matthew-reynolds: Thanks for clarifying that, I will remove the TODO.

transmission_interface/package.xml
matthew-reynolds: Dependencies revised again in 78c1957. I have one main question coming out of this:

Do we need to `<exec_depend>` on header-only packages containing plugin interfaces (ie. `hardware_interface`), or are `<build_depend>` and `<build_export_depend>` sufficient?

I don't know why the `exec_` would be required, but [the pluginlib wiki](http://wiki.ros.org/pluginlib#pluginlib.2BAC8-pluginlib_groovy.Registering_Plugin_with_ROS_Package_System) states that "the [plugin-]providing package must depend directly on the package containing the plugin interface", with the code example containing both `run_` and `build_` dependencies.

I don't know whether the wiki's advice applies to header-only packages or not, so for now I've simply used `<depend>` as a catch-all. I suspect this is unnecessary, but I preferred to err on the side of caution since I'm unsure.

---

I know this PR is getting long, sorry for dragging it out. I've kinda fallen into the trap of tackling too much in a single PR. Unless you've got insight to resolve this here, I will leave it as is and open a follow-up issue.

hardware_interface/CMakeLists.txt
matthew-reynolds: TODO removed in def3541

controller_manager/CMakeLists.txt
matthew-reynolds: Fixed in def3541

bmagyar (approved):

bmagyar:
Thanks @matthew-reynolds for the great for and @gavanderhoorn for the additional review, I am merging this now!

transmission_interface/package.xml
matthew-reynolds: Follow-up in #414

controller_manager_tests/CMakeLists.txt
```diff
@@ -2,56 +2,77 @@ cmake_minimum_required(VERSION 2.8.3)
...
+    src/extensible_controllers.cpp
+    src/my_dummy_controller.cpp
+    src/my_robot_hw.cpp
+    src/pos_eff_controller.cpp
+    src/pos_eff_opt_controller.cpp
+    src/vel_eff_controller.cpp
+  )
+  target_link_libraries(${PROJECT_NAME} ${catkin_LIBRARIES})
+
+  add_executable(dummy_app EXCLUDE_FROM_ALL src/dummy_app.cpp)
```
mathias-luedtke, line 40: Where did `EXCLUDE_FROM_ALL` come from?

matthew-reynolds: When looking at the targets in all the test packages in https://github.com/ros/ros_comm/tree/melodic-devel/test as an example, I saw they marked targets intended for tests as `EXCLUDE_FROM_ALL`. Similarly, targets generated by `catkin_add_gtest()` and `catkin_add_gmock()` enable the property.

I admit, I don't fully understand why this is required, since the test targets are already wrapped in the `if(CATKIN_ENABLE_TESTING)`. I have the impression it was to cover an edge-case with `catkin_make` (as opposed to `catkin_make_isolated` or `catkin_tools`), something to do with the single CMake invocation over the entire workspace, but instead of properly researching and understanding it, I just used those packages as examples and mimicked.

mathias-luedtke: I think you have to add the targets to `tests` manually:
https://github.com/ros/ros_comm/blob/8d9af410949423342150b79e763348f963e19734/test/test_roscpp/test/src/CMakeLists.txt#L226

mathias-luedtke: Or use `add_rostest_gtest`, which would [automate](https://github.com/ros/ros_comm/blob/melodic-devel/tools/rostest/cmake/rostest-extras.cmake.em#L67) most of the steps.

matthew-reynolds: > I think you have to add the targets to `tests` manually

Ah, I think you're right, that makes sense. Since those targets are only depended on by targets in `add_rostest`, not `add_rostest_gtest`, they aren't directly or transitively depended on by the `tests` target, and since I added the `EXCLUDE_FROM_ALL` property, they aren't part of the `all` target either. Sounds like this may be the cause of [the colcon failure you mentioned](https://github.com/ros-controls/ros_control/issues/410#issuecomment-574397248).

I will re-add the `add_dependencies(tests target_name)` for executable targets used in tests. What are your and @bmagyar's thoughts on also adding library targets as dependencies? Back to the `ros_comm` test package examples, they seem to only mark executables as deps, and rely on libraries being pulled in transitively.

mathias-luedtke: > rely on libraries being pulled in transitively.

This is sufficient. Add direct dependencies only.

Issue #403, opened 2019-12-27
Modernize the codebase using C++11/14 features


bmagyar:
A non-comprehensive list of TODOs:
* [x] replace `boost` with `std` whenever possible
* [x] replace iterator and `BOOST_FOREACH` with range-based `for` whenever possible
* [x] replace `enum`s with `class enum`s whenever possible
* [x] use `override` in derived functions
* [x] watch out if you can add more const!
* [x] use `nullptr` instead of good ol' `0`
* [x] use `#pragma once` in all headers

bmagyar:
FYI @matthew-reynolds please feel free to add your ideas on top!

matthew-reynolds:
Couple of other thoughts:

 - [ ] use `= default` and `= delete` where appropriate
 - [ ] use `noexcept` where appropriate
 - [ ] use `final` where appropriate
 - [ ] replace `typedef` with `using`
 - [ ] use `auto` when it improves readability (e.g. rather than spelling out the exact iterator when iterating over stl containers, etc)

And some low-priority/more controversial changes:
 - [ ] use pass-by-value and `move` where appropriate (e.g. rather than pass-by-const-ref for some parameters in constructors)
 - [ ] use default member initializers (brace-or-equal-initializer) for member variables to simplify constructors (rather than listing out all the 0s and nullptrs in every constructor)

matthew-reynolds:
As mentioned [here](https://github.com/ros-controls/ros_control/pull/395/files#r362134419), another potential cleanup:
 - [ ] create a `.clang-format` and apply it to the whole codebase
 - [ ] perhaps create a `.clang-tidy` and apply it the whole codebase

(I suppose it might fall a bit outside the scope of "modernizing" since it's more "cleanup". If preferred, I can open a separate issue to track this).

bmagyar:
>   (I suppose it might fall a bit outside the scope of "modernizing" since it's more "cleanup". If preferred, I can open a separate issue to track this).

Indeed I have split this into a separate issue here: https://github.com/ros-controls/ros_control/issues/407

suab321321:
@matthew-reynolds I would like to contribute

matthew-reynolds:
@suab321321 Thanks, I've got PRs either open or in-progress on my end for almost all of these tasks, except for going through with a fine-toothed comb and adding `const`s and, where it improves readability, `auto`. Would certainly appreciate help in this front. I would also appreciate someone looking into where we can leverage move semantics.

On a related note:

[...]

PR #406, opened 2019-12-30
Use #pragma once

matthew-reynolds:
Part of #403.

Replace all preprocessor header guards with `#pragma once`. Also adds missing header guard to `transmission_interface/test/loader_utils.h`. Every `.h` file now includes `#pragma once`.

### DIFF
--- a/combined_robot_hw/include/combined_robot_hw/combined_robot_hw.h
+++ b/combined_robot_hw/include/combined_robot_hw/combined_robot_hw.h
@@ -25,8 +25,8 @@
 // POSSIBILITY OF SUCH DAMAGE.
 //////////////////////////////////////////////////////////////////////////////
 
-#ifndef COMBINED_ROBOT_HW_COMBINED_ROBOT_HW_H
-#define COMBINED_ROBOT_HW_COMBINED_ROBOT_HW_H
+#pragma once
+
 
 #include <list>
 #include <map>
@@ -117,5 +117,3 @@ protected:
 };
 
 }
-
-#endif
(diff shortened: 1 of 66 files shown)


Karsten1987:
I just wanted to mention here that this pragma is not standard compliant. As most compilers do support it, but as said it’s not in the official standard. 
https://stackoverflow.com/questions/23696115/is-pragma-once-part-of-the-c11-standard#23699893

bmagyar:
If all relevant compilers support it (heck, even MSVC does!) then I think it is a safe bet.
supported compilers

matthew-reynolds:
Rebased and updated following #395

bmagyar (approved):
Looks great, thanks a lot!

mathias-luedtke:
Ah, I am too late for this one.

> If all relevant compilers support it (heck, even MSVC does!) then I think it is a safe bet.

Not all versions support it and some implementations have serious flaws. 
Even worse: If a compiler does not support it, it must (?) ignore it.

mathias-luedtke:
I would stick to the guidelines: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#sf8-use-include-guards-for-all-h-files

matthew-reynolds:
I don't have a strong opinion here, I'm just following #403. Happy to revert to #ifndef guards if that's the consensus of you maintainers.

I do wonder how many C++14-compatible compilers don't support `#pragma once`, but that's not an easy audit to make to cover every obscure compiler out there.

PR #412, opened 2020-01-12
[Noetic] Replace enums with enum classes

matthew-reynolds:
Part of #403

Heads up, this is a breaking change. Might be better off targetting `noetic-devel`.

### DIFF
--- a/controller_interface/include/controller_interface/controller.h
+++ b/controller_interface/include/controller_interface/controller.h
@@ -52,7 +52,7 @@ template <class T>
 class Controller: public virtual ControllerBase
 {
 public:
-  Controller() {}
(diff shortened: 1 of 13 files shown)


bmagyar:
Breaking indeed but I think this is mostly internal. What do you think, @ipa-mdl , have you encountered 3rd party code that'd depend a lot on these old-fashioned enums?

mathias-luedtke:
I would suggest to not change enums in melodic.
There is no strong need for this change..

bmagyar:
Let's keep this here for a bit more if you don't mind, then branch for Noetic

bmagyar (approved):

PR #413, opened 2020-01-13
Use range-based for loop

matthew-reynolds:
Part of #403

### DIFF
--- a/hardware_interface/include/hardware_interface/internal/interface_manager.h
+++ b/hardware_interface/include/hardware_interface/internal/interface_manager.h
@@ -54,10 +54,8 @@ struct CheckIsResourceManager {
   template <typename C>
   static void callCM(typename std::vector<C*>& managers, C* result, typename C::resource_manager_type*)
   {
-    std::vector<typename C::resource_manager_type*> managers_in;
-    // we have to typecase back to base class
-    for(typename std::vector<C*>::iterator it = managers.begin(); it != managers.end(); ++it)
-      managers_in.push_back(static_cast<typename C::resource_manager_type*>(*it));
+    // We have to typecase back to base class
+    std::vector<typename C::resource_manager_type*> managers_in(managers.begin(), managers.end());
     C::concatManagers(managers_in, result);
   }
 
@@ -171,9 +169,8 @@ public:
     }
 
     // look for interfaces registered in the registered hardware
-    for(InterfaceManagerVector::iterator it = interface_managers_.begin();
-        it != interface_managers_.end(); ++it) {
-      T* iface = (*it)->get<T>();
+    for (const auto& interface_manager : interface_managers_) {
+      T* iface = interface_manager->get<T>();
       if (iface)
         iface_list.push_back(iface);
     }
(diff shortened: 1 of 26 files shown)


hardware_interface/include/hardware_interface/internal/interface_manager.h
```diff
@@ -54,10 +54,8 @@ struct CheckIsResourceManager {
   template <typename C>
   static void callCM(typename std::vector<C*>& managers, C* result, typename C::resource_manager_type*)
   {
-    std::vector<typename C::resource_manager_type*> managers_in;
-    // we have to typecase back to base class
-    for(typename std::vector<C*>::iterator it = managers.begin(); it != managers.end(); ++it)
-      managers_in.push_back(static_cast<typename C::resource_manager_type*>(*it));
+    // We have to typecase back to base class
```
bmagyar: I think the comment here was meant to say
```suggestion
    // We have to typecast back to base class
```

hardware_interface/include/hardware_interface/internal/resource_manager.h
```diff
@@ -80,9 +80,9 @@ class ResourceManager : public ResourceManagerBase
   {
     std::vector<std::string> out;
     out.reserve(resource_map_.size());
-    for(typename ResourceMap::const_iterator it = resource_map_.begin(); it != resource_map_.end(); ++it)
+    for (const auto& pair : resource_map_)
```
bmagyar: When `range-based for`-ing a `map` I typically like to give the pair a very verbose name, in this case it could be something like
```suggestion
    for (const auto& name_and_joint_handle : resource_map_)
```

This helps understanding what `.first` and `.second` refers to down the line. The common argument against this is that we pair naming with the semantics of that map but I'd argue that's kind of the whole point of programming languages more advanced than bytecode, we can name things to convey context.

hardware_interface/include/hardware_interface/robot_hw.h
```diff
@@ -92,38 +92,31 @@ class RobotHW : public InterfaceManager
...
-          resource_map[*resource_it].push_back(*info_it);
+          resource_map[iface_resource].push_back(controller_info);
         }
       }
     }
 
     // Enforce resource exclusivity policy: No resource can be claimed by more than one controller
     bool in_conflict = false;
-    for (CtrlInfoMap::iterator it = resource_map.begin(); it != resource_map.end(); ++it)
+    for (const auto& pair : resource_map)
```
bmagyar: following the above, this could be

```suggestion
    for (const auto& resource_name_and_claiming_controllers : resource_map)
```

bmagyar (changes requested):

hardware_interface/include/hardware_interface/internal/interface_manager.h
matthew-reynolds: Updated in 60b2f95

hardware_interface/include/hardware_interface/internal/resource_manager.h
matthew-reynolds: Sounds good to me, will update the names to be more meaningful.

hardware_interface/include/hardware_interface/robot_hw.h
matthew-reynolds: Done in 94e3092

hardware_interface/include/hardware_interface/internal/resource_manager.h
matthew-reynolds: Done in 94e3092, though I used a slightly more generic name

bmagyar (approved):
Thanks very much for this!

Issue #410, opened 2020-01-12
Investigate running tests out of the devel space


matthew-reynolds:
Following up from https://github.com/ros-controls/ros_control/pull/404#discussion_r364924536:

> Looks like tests fail if we don't install plugins. Will investigate.

_Originally posted by @matthew-reynolds_

This seems that the way `industrial_ci` and `catkin_tools` work together causes us to require installing libraries and plugin `.xml` files, even if they're only used for tests.

Here is a summary of my understanding:

 - `industrial_ci` calls `catkin config --install` before building or running any tests. This is a good thing for building, but causes some hiccups when running tests.
 - `catkin_tools`, when running tests, [doesn't properly source the **devel** space](https://github.com/catkin/catkin_tools/issues/397)
 - This combination of behaviours results in test resources not being found unless they're installed. However, since these resources are only for testing, it isn't ideal to be installing them.

For now, we're installing the required libraries and the plugin `.xml` files. Until `catkin_tools` is updated or unless `industrial_ci` implements a workaround, I don't think there's much we can do to avoid these installs.

I'm mainly opening this issue to track the behaviour and watch for upstream changes, but if there's a better solution that I've missed, I'd love to hear it.

mathias-luedtke:
The devel space should not be used, and in general all plugin files must be installed.
It looks like #404 does not work with `colcon` anymore (https://travis-ci.com/ros-industrial/industrial_ci/jobs/275277364)

matthew-reynolds:
> The devel space should not be used

Interesting, can you elaborate on that? I got the impression from that issue that it would be preferable to use the devel space, if possible. My understanding is that running tests out of the install space has the benefit of testing that files are actually installed correctly, but the disadvantage of requiring test resources and files be installed

> It looks like #404 does not work with colcon anymore

Looks like this is due to the test nodes not being installed anymore. We ended up removing the installs for non-pluginlib-related files, sounds like that was not a good idea. Any idea why this works with `catkin_make` and `catkin_tools` but not `colcon`? I haven't played with `colcon` yet and don't know how it differs under the hood.

mathias-luedtke:
> > The devel space should not be used

> Interesting, can you elaborate on that?

The devel space was mostly a hack, and covers a lot of problems.
That's why it got removed in ROS2. In addition, `colcon` does not support it anymore.

> Looks like this is due to the test nodes not being installed anymore. 

The tests do not have to get installed necessarily, because they can get started from the build space (?).
[...]

PR #415, opened 2020-01-14
Add test nodes to CMake target "tests"

matthew-reynolds:
This resolves the `colcon` test failure brought up by @ipa-mdl here: https://github.com/ros-controls/ros_control/issues/410#issuecomment-574397248

See https://github.com/ros-controls/ros_control/pull/404#discussion_r366604167 for full discussion on this change. In short, targets have either got to be part of the `all` or `tests` targets, and #404 removed some targets from `all` without adding them to `tests`.

### DIFF
--- a/combined_robot_hw_tests/CMakeLists.txt
+++ b/combined_robot_hw_tests/CMakeLists.txt
@@ -36,6 +36,7 @@ if(CATKIN_ENABLE_TESTING)
 
   add_executable(combined_robot_hw_dummy_app EXCLUDE_FROM_ALL src/dummy_app.cpp)
   target_link_libraries(combined_robot_hw_dummy_app ${PROJECT_NAME} ${catkin_LIBRARIES})
+  add_dependencies(tests combined_robot_hw_dummy_app)
 
   add_rostest_gtest(combined_robot_hw_test
     test/combined_robot_hw_test.test
@@ -56,8 +57,7 @@ endif()
 ## Install ##
 #############
 
-# NOTE: Libraries and plugins required for tests are installed since CI
-# runs tests out of the install space rather than the devel space
+# NOTE: Plugins required for tests must be installed
 
 if(CATKIN_ENABLE_TESTING)
 
--- a/combined_robot_hw_tests/CMakeLists.txt
+++ b/combined_robot_hw_tests/CMakeLists.txt
@@ -48,7 +48,7 @@ if(CATKIN_ENABLE_TESTING)
     test/cm_test.test
     test/cm_test.cpp
   )
-  add_dependencies(combined_robot_hw_cm_test combined_robot_hw_dummy_app ${catkin_EXPORTED_TARGETS})
+  add_dependencies(combined_robot_hw_cm_test ${catkin_EXPORTED_TARGETS})
   target_link_libraries(combined_robot_hw_cm_test ${PROJECT_NAME} ${catkin_LIBRARIES})
 endif()
 
--- a/controller_manager_tests/CMakeLists.txt
+++ b/controller_manager_tests/CMakeLists.txt
@@ -40,10 +40,12 @@ if(CATKIN_ENABLE_TESTING)
   add_executable(dummy_app EXCLUDE_FROM_ALL src/dummy_app.cpp)
   add_dependencies(dummy_app ${catkin_EXPORTED_TARGETS})
   target_link_libraries(dummy_app ${PROJECT_NAME} ${catkin_LIBRARIES})
+  add_dependencies(tests dummy_app)
 
   add_executable(cm_test EXCLUDE_FROM_ALL test/cm_test.cpp)
   add_dependencies(cm_test ${catkin_EXPORTED_TARGETS})
   target_link_libraries(cm_test ${catkin_LIBRARIES} ${GTEST_LIBRARIES})
+  add_dependencies(tests cm_test)
 
   catkin_add_nosetests(test)
 
@@ -59,8 +61,7 @@ endif()
 ## Install ##
 #############
 
-# NOTE: Libraries and plugins required for tests are installed since CI
-# runs tests out of the install space rather than the devel space
+# NOTE: Plugins required for tests must be installed
 
 if(CATKIN_ENABLE_TESTING)
 
--- a/controller_manager_tests/CMakeLists.txt
+++ b/controller_manager_tests/CMakeLists.txt
@@ -49,11 +49,11 @@ if(CATKIN_ENABLE_TESTING)
 
   catkin_add_nosetests(test)
 
-  add_rostest(test/cm_test.test DEPENDENCIES dummy_app cm_test)
+  add_rostest(test/cm_test.test)
 
   add_rostest(test/cm_msgs_utils_rostest.test)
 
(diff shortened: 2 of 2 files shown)


combined_robot_hw_tests/CMakeLists.txt
```diff
@@ -36,6 +36,7 @@ if(CATKIN_ENABLE_TESTING)
 
   add_executable(combined_robot_hw_dummy_app EXCLUDE_FROM_ALL src/dummy_app.cpp)
   target_link_libraries(combined_robot_hw_dummy_app ${PROJECT_NAME} ${catkin_LIBRARIES})
+  add_dependencies(tests combined_robot_hw_dummy_app)
```
mathias-luedtke, line 39: I guess this should better be wrapped with `if(TARGET tests)` as in `ros_comm`.

mathias-luedtke (changes requested):
Thanks!

combined_robot_hw_tests/CMakeLists.txt
bmagyar: I'd rather move it into the `if(CATKIN_ENABLE_TESTING)` section, I strongly dislike generating more cmake blurb

mathias-luedtke: > I'd rather move it into the if(CATKIN_ENABLE_TESTING)

It already is.

> I strongly dislike generating more cmake blurb

I don't understand why we moved to `ros_comm` style in the first place.
A simpler alternative is to just remove `EXCLUDE_FROM_ALL`.
The packages in question are test packages anyway.

matthew-reynolds: >A simpler alternative is to just remove EXCLUDE_FROM_ALL.
The packages in question are test packages anyway.

True, but I think the extra safeguards against installing test targets is worthwhile.

So, the conclusion here is to add the `if(TARGET tests)`?

matthew-reynolds: A final note: `CATKIN_ENABLE_TESTING` defaults to `ON`, so the `EXCLUDE_FROM_ALL` is quite useful for speeding up regular builds and only building tests when appropriate.

---

Back to the `if(TARGET tests)`:

> So, the conclusion here is to add the if(TARGET tests)?

I dug into this more and changed my mind, I don't think we should add this.

I don't know why `ros_comm` uses it around their `add_dependencies(tests ...)`. The Catkin CMake library will _always_ create the `tests` target ([tests.cmake L81](https://github.com/ros/catkin/blob/b6410a5257bb991481b28570899830704d768979/cmake/test/tests.cmake#L81)) unless `CATKIN_ENABLE_TESTING` is set to `OFF`. That same file also implies that it is sufficient to use only the `CATKIN_ENABLE_TESTING` check for test stuff ([tests.cmake L34](https://github.com/ros/catkin/blob/b6410a5257bb991481b28570899830704d768979/cmake/test/tests.cmake#L34)).

Additionally, the `catkin_add_gtest()` and `_gmock()` functions don't do this check before they `add_dependencies(tests ...)`; They only check for `CATKIN_ENABLE_TESTING`, and issue a warning if they are called while `CATKIN_ENABLE_TESTING` is `OFF`.

Since we are already within `if(CATKIN_ENABLE_TESTING)`, I think we are safe without `if(TARGET tests)`. If we want more clarity on why `ros_comm` uses it, we could just ask the `ros_comm` maintainers why they use the check and make our decision accordingly.

matthew-reynolds: In fact, it looks like `ros_comm` even uses `if(TARGET ...)` for targets they just created the line before... See [test_rosbag L37](https://github.com/ros/ros_comm/blob/902fb00fc7b2e881575a270a99c3a077ba5cdbba/test/test_rosbag/CMakeLists.txt#L37). So their usage isn't specific to the `tests` target.

mathias-luedtke: > True, but I think the extra safeguards against installing test targets is worthwhile.

We are talking about `*_test`  packages, their only purpose is to provide everything for the tests.

> Additionally, the catkin_add_gtest() and _gmock() functions don't do this check before they add_dependencies(tests ...)

[They do it for executables](https://github.com/ros/catkin/blob/dfc93c72d9799f57442d08c2f6e9fd494bba95c1/cmake/test/gtest.cmake#L82)

mathias-luedtke: > In fact, it looks like ros_comm even uses if(TARGET ...) for targets they just created the line before

Yes, I guess it is best practice to check if the *external* target exists before using it.

mathias-luedtke (changes requested):
Please remove fc28b088a1e1ee107a1608eb3c01d7de2fd2bb21

The dependency is needed to determine the correct order, e.g. for parallel execution.
In addition the `tests`target does not get built automatically, if I am not mistaken.

combined_robot_hw_tests/CMakeLists.txt
matthew-reynolds: > We are talking about *_test packages, their only purpose is to provide everything for the tests.

Indeed, but all the better if they can do so while installing as few artifacts as possible (ie. only plugins, which are required to be installed by pluginlib).

Also, while we may disagree on this point, I think the build speedup and more expected behaviour (only building these targets in test builds) still makes this worthwhile.

> [They do it for executables](https://github.com/ros/catkin/blob/dfc93c72d9799f57442d08c2f6e9fd494bba95c1/cmake/test/gtest.cmake#L82)

This line is checking that the newly created target exists (Similar to [test_rosbag L37](https://github.com/ros/ros_comm/blob/902fb00fc7b2e881575a270a99c3a077ba5cdbba/test/test_rosbag/CMakeLists.txt#L37)), not checking that the `tests` target exists, as we're debating.

matthew-reynolds:
>Please remove fc28b08

These targets are all run-time dependencies, not build-time, and therefore shouldn't be specified as such. From the commit msg:
```
All "tests" targets will be built before any tests are run. These explicit
dependencies are neither necessary nor standard/conventional.
```
Additionally, those build-time dependencies were only introduced in the #404 refactor - none of `ros_comm`, `ros_controllers`, or `ros_control` before #404 have it.

>The dependency is needed to determine the correct order, e.g. for parallel execution.

Every test depends on the entire `tests` target ([tests.cmake L173](https://github.com/ros/catkin/blob/b6410a5257bb991481b28570899830704d768979/cmake/test/tests.cmake#L173)) effectively forcing all tests to build before running any of them. So I don't think order and parallelization are concerns here.

> In addition the `tests` target does not get built automatically, if I am not mistaken.

The `tests` target is a dependency of the `run_tests` target, so every `run_tests` build will build these nodes. They won't be built during a normal build though, which IMO is the correct behaviour.

matthew-reynolds:
@ipa-mdl @bmagyar gentle bump. I'd like to fix up `colcon` builds and tie up the loose ends from #404.

matthew-reynolds:
Another ping @bmagyar @ipa-mdl, let's get this finished up, so I can make the mirror whatever we decide on over to `ros_controllers`.

bmagyar (approved):

PR #416, opened 2020-01-20
Improve controller and resource filtering for CombinedRobotHW

toliver:
This PR improves the filtering that CombinedRobotHW does on controllers, interfaces and resources before it passes the `start_list` and `stop_list` to the underlying `RobotHW`s.
So now:
- If none of the resources in a certain interface (that had some resources) belong to a certain RobotHW instance, then that interface will be filtered out (for that RobotHW instance).
- If none of the resources in the interfaces of a controller belong to a certain RobotHW instance, then that controller will be filtered out (for that RobotHW instance).

Test cases have been added to cover these two cases.

### DIFF
--- a/combined_robot_hw/src/combined_robot_hw.cpp
+++ b/combined_robot_hw/src/combined_robot_hw.cpp
@@ -243,10 +243,16 @@ namespace combined_robot_hw
             filtered_resources.insert(*ctrl_res);
           }
         }
-        filtered_iface_resources.resources = filtered_resources;
-        filtered_controller.claimed_resources.push_back(filtered_iface_resources);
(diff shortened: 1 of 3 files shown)


bmagyar:
Is this something you'd strongly need in Kinetic or happy to have it only in Melodic and above?

toliver:
It would be good to have it on Kinetic as well if it is not a pain for you.
We can prepare another PR for Melodic if that helps.

bmagyar:
yes please create one for melodic too

toliver:
Done. Melodic PR in https://github.com/ros-controls/ros_control/pull/422

bmagyar (approved):

PR #417, opened 2020-01-23
Use auto keyword

suab321321:
Signed-off-by: AbhinavSingh <singhabhinav9051571833@gmail.com>

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -80,18 +80,18 @@ void ControllerManager::update(const ros::Time& time, const ros::Duration& perio
 
   // Restart all running controllers if motors are re-enabled
   if (reset_controllers){
-    for (size_t i=0; i<controllers.size(); i++){
-      if (controllers[i].c->isRunning()){
-        controllers[i].c->stopRequest(time);
-        controllers[i].c->startRequest(time);
+    for (const auto& controller : controllers_lists_[used_by_realtime_]){
+      if (controller.c->isRunning()){
+        controller.c->stopRequest(time);
+        controller.c->startRequest(time);
       }
     }
   }
 
 
   // Update all controllers
-  for (size_t i=0; i<controllers.size(); i++)
-    controllers[i].c->updateRequest(time, period);
+  for (const auto& controller : controllers_lists_[used_by_realtime_])
+    controller.c->updateRequest(time, period);
 
   // there are controllers to start/stop
   if (switch_params_.do_switch)
@@ -105,11 +105,10 @@ controller_interface::ControllerBase* ControllerManager::getControllerByName(con
   // Lock recursive mutex in this context
   boost::recursive_mutex::scoped_lock guard(controllers_lock_);
 
-  auto &controllers = controllers_lists_[current_controllers_list_];
-  for (size_t i = 0; i < controllers.size(); ++i)
+  for (const auto& controller : controllers_lists_[current_controllers_list_])
   {
-    if (controllers[i].info.name == name)
-      return controllers[i].c.get();
+    if (controller.info.name == name)
+      return controller.c.get();
   }
   return nullptr;
 }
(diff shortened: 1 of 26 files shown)


suab321321:
#403 @matthew-reynolds I have made iterators auto

matthew-reynolds:
This PR seems to do a lot of for loop stuff, which is already done in #413. Can you remove those changes, but leave your other changes to be reviewed?

suab321321:
@matthew-reynolds I have made the `iterator` auto which is used for both `for loop` and also for `iterators` pointing to `std::find()` results.Should I make the for loop auto iterator to default ones,beacuse in some files`auto` was not used and I thought it will good to make them auto.

bmagyar:
#413 is now merged and as a result, this PR is conflicting.

@suab321321 , for a smooth, collaborative experience, it is good to avoid modifying code that is already under review on a different PR. As @matthew-reynolds raised, you did some duplicate work that is not causing trouble on this PR. Please resolve these conflicts.

suab321321:
@matthew-reynolds @bmagyar will update soon

suab321321:
@matthew-reynolds @bmagyar is it okay?

controller_manager/src/controller_manager.cpp
```diff
@@ -76,6 +76,7 @@ ControllerManager::~ControllerManager()
 void ControllerManager::update(const ros::Time& time, const ros::Duration& period, bool reset_controllers)
 {
   used_by_realtime_ = current_controllers_list_;
+  auto &controllers = controllers_lists_[used_by_realtime_];
```
bmagyar: I don't see this variable used here, can probably be removed

controller_manager/src/controller_manager.cpp
```diff
@@ -294,15 +295,13 @@ bool ControllerManager::loadController(const std::string& name)
     try
     {
       // Trying loading the controller using all of our controller loaders. Exit once we've found the first valid loaded controller
-      std::list<ControllerLoaderInterfaceSharedPtr>::iterator it = controller_loaders_.begin();
-      while (!c && it != controller_loaders_.end())
+      for (auto const& it : controller_loaders_)
```
bmagyar: Careful here. Read the above comment carefully. I agree that this part could be improved but I am unsure at the moment how best to achieve that. Your suggested change alters the algorithm

controller_manager/src/controller_manager.cpp
```diff
@@ -523,7 +522,7 @@ bool ControllerManager::switchController(const std::vector<std::string>& start_c
   switch_start_list_.clear();
   switch_stop_list_.clear();
 
-  std::vector<ControllerSpec> &controllers = controllers_lists_[current_controllers_list_];
+  auto &controllers = controllers_lists_[current_controllers_list_];
```
bmagyar: This could probably be const

controller_manager/src/controller_manager.cpp
```diff
@@ -391,7 +390,7 @@ bool ControllerManager::unloadController(const std::string &name)
     }
     std::this_thread::sleep_for(std::chrono::microseconds(200));
   }
-  std::vector<ControllerSpec>
+  auto
```
bmagyar: I would argue that `auto` here actually hides useful information

bmagyar (changes requested):

controller_manager/src/controller_manager.cpp
suab321321:
```
for(const auto& it:controller_loaders_.begin()){
   if(c==controller_loaders_.end())
      break;
   for(const auto& cur_type:(it)->getDeclaredClass()){
      if(cur_type==type)
         c=(it)->createInstance(type);
      else
            c=it;
   }
}
```

suab321321: @bmagyar will this do?

suab321321: @bmagyar sorry I m utterly wrong did not read the code correctly.

suab321321: @bmagyar sir is my current implementation okay?

bmagyar: Let's settle for a simpler solution, auto the iterator and leave the rest as-is.

`std::list<ControllerLoaderInterfaceSharedPtr>::iterator it = controller_loaders_.begin();` -> `auto it = controller_loaders_.begin();`

bmagyar (changes requested):

suab321321:
@bmagyar

bmagyar (approved):
Thanks for sticking through this review!

PR #419, opened 2020-01-27
Replace internal::is_permutation with std

matthew-reynolds:
Somewhat related to #403.

Since Melodic requires >=C++14, we can replace this custom `is_permutation` with C++11's `std::is_permutation`.

According to http://www.cplusplus.com/reference/algorithm/is_permutation, the standard is equivalent to what Adolfo wrote 6 years ago, so this shouldn't change any behaviour.


bmagyar (approved):
Thanks for this!

PR #421, opened 2020-01-28
Use default member initializers

matthew-reynolds:
Part of #403.

When possible, prefer the use of default member initializers. Note that I didn't make an effort to add default initializers for all members, I only focussed on moving initializations out of constructors.

This sets the stage for cleaning up constructors using `= default` and `= delete`. (Follow-up PR containing those changes is ready, I'll rebase and open it once this PR is completed)

### DIFF
--- a/controller_manager/include/controller_manager/controller_manager.h
+++ b/controller_manager/include/controller_manager/controller_manager.h
@@ -191,24 +191,14 @@ private:
 
   struct SwitchParams
   {
-    SwitchParams()
-      : do_switch(false)
-      , started(false)
-      , init_time(ros::TIME_MAX)
-      , strictness(0)
-      , start_asap(false)
-      , timeout(0.0)
-    {
-    }
+    bool do_switch      = false;
+    bool started        = false;
+    ros::Time init_time = ros::TIME_MAX;
 
-    bool do_switch;
-    bool started;
-    ros::Time init_time;
-
-    // switch options
-    int strictness;
-    bool start_asap;
-    double timeout;
+    // Switch options
+    int strictness  = 0;
+    bool start_asap = false;
+    double timeout  = 0.0;
   };
 
   SwitchParams switch_params_;
@@ -224,9 +214,9 @@ private:
   /// Double-buffered controllers list
   std::vector<ControllerSpec> controllers_lists_[2];
   /// The index of the current controllers list
-  int current_controllers_list_;
+  int current_controllers_list_ = 0;
   /// The index of the controllers list being used in the real-time thread.
-  int used_by_realtime_;
+  int used_by_realtime_ = -1;
   /*\}*/
 
 
--- a/controller_manager/include/controller_manager/controller_manager.h
+++ b/controller_manager/include/controller_manager/controller_manager.h
@@ -191,14 +191,14 @@ private:
 
   struct SwitchParams
   {
-    bool do_switch      = false;
-    bool started        = false;
-    ros::Time init_time = ros::TIME_MAX;
+    bool do_switch      = {false};
+    bool started        = {false};
+    ros::Time init_time = {ros::TIME_MAX};
 
     // Switch options
-    int strictness  = 0;
-    bool start_asap = false;
-    double timeout  = 0.0;
+    int strictness  = {0};
+    bool start_asap = {false};
+    double timeout  = {0.0};
   };
 
   SwitchParams switch_params_;
@@ -214,9 +214,9 @@ private:
   /// Double-buffered controllers list
   std::vector<ControllerSpec> controllers_lists_[2];
   /// The index of the current controllers list
-  int current_controllers_list_ = 0;
+  int current_controllers_list_ = {0};
   /// The index of the controllers list being used in the real-time thread.
-  int used_by_realtime_ = -1;
+  int used_by_realtime_ = {-1};
   /*\}*/
 
 
--- a/hardware_interface/include/hardware_interface/actuator_command_interface.h
+++ b/hardware_interface/include/hardware_interface/actuator_command_interface.h
@@ -61,7 +61,7 @@ public:
   double* getCommandPtr() {return cmd_;}
 
 private:
-  double* cmd_ = nullptr;
+  double* cmd_ = {nullptr};
 };
 
 /** \brief Hardware interface to support commanding an array of actuators.
--- a/hardware_interface/include/hardware_interface/actuator_command_interface.h
+++ b/hardware_interface/include/hardware_interface/actuator_command_interface.h
@@ -40,7 +40,7 @@ namespace hardware_interface
 class ActuatorHandle : public ActuatorStateHandle
 {
 public:
-  ActuatorHandle() : ActuatorStateHandle(), cmd_(nullptr) {}
+  ActuatorHandle() {}
 
   /**
    * \param as This actuator's state handle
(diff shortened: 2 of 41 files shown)


controller_manager/include/controller_manager/controller_manager.h
```diff
@@ -224,9 +214,9 @@ class ControllerManager{
   /// Double-buffered controllers list
   std::vector<ControllerSpec> controllers_lists_[2];
   /// The index of the current controllers list
-  int current_controllers_list_;
+  int current_controllers_list_ = 0;
```
piraka9011: Consider using brace-initialization here.
```suggestion
  int current_controllers_list_{0};
```

controller_manager/include/controller_manager/controller_manager.h
```diff
@@ -224,9 +214,9 @@ class ControllerManager{
...
   /// The index of the controllers list being used in the real-time thread.
-  int used_by_realtime_;
+  int used_by_realtime_ = -1;
```
piraka9011:
```suggestion
  int used_by_realtime_{-1};
```

hardware_interface/include/hardware_interface/actuator_command_interface.h
```diff
@@ -61,7 +61,7 @@ class ActuatorHandle : public ActuatorStateHandle
   double* getCommandPtr() {return cmd_;}
 
 private:
-  double* cmd_;
+  double* cmd_ = nullptr;
```
piraka9011:
```suggestion
  double* cmd_{nullptr};
```

piraka9011:
Thanks for the PR!
If you're going to change the default initialization of variables then I strongly suggest you use brace initialization ([quick read](https://blog.quasardb.net/2017/03/05/cpp-braced-initialization) on why).

matthew-reynolds:
>I strongly suggest you use brace initialization

I chose equals rather than brace initialization based on @bmagyar's comments on #403 (Specifically [here](https://github.com/ros-controls/ros_control/issues/403#issuecomment-575144402) and [here](https://github.com/ros-controls/ros_control/issues/403#issuecomment-578977097)).

I don't have a strong opinion here, and the [C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c48-prefer-in-class-initializers-to-member-initializers-in-constructors-for-constant-initializers) don't state a preference. Happy to change to brace, though I'd like to hear from @bmagyar first since he asked for equals.

piraka9011:
> I don't have a strong opinion here, and the [C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c48-prefer-in-class-initializers-to-member-initializers-in-constructors-for-constant-initializers) don't state a preference. Happy to change to brace, though I'd like to hear from @bmagyar first since he asked for equals.

While not stated directly, see Sutter's [post](https://herbsutter.com/2013/05/09/gotw-1-solution/) (He's the editor of the ISO guidelines :smiley: ). 

> **Guideline:** Prefer to use initialization with { }, such as vector<int> v = { 1, 2, 3, 4 }; or auto v = vector<int>{ 1, 2, 3, 4 };, because it’s more consistent, more correct, and avoids having to know about old-style pitfalls at all. In single-argument cases where you prefer to see only the = sign, such as int i = 42; and auto x = anything; omitting the braces is fine. … 

See also his post [Elements of Modern C++ Style](https://herbsutter.com/elements-of-modern-c-style/).
@bmagyar while the equals approach is 'simple', brace initialization:

> ...always initializes the variable, and is never ambiguous with a function declaration. No vex, no fuss, no muss. - [Sutter](https://herbsutter.com/2013/05/09/gotw-1-solution/)

matthew-reynolds:
@bmagyar gentle bump. Unless you've got a strong reason to prefer equals, I will switch over to brace.

bmagyar:
Thanks for those posts. I am fully onboard with applying new language features where they make life easier and the code more concise *however* I'm also questioning every case where I see new styles applied without common sense.

In the posts referenced above I found two where I'm seeing recommendations for the opposite to what you are saying here, specifically for the single argument simple case.

> What hasn’t changed: When initializing a local variable whose type is non-POD or auto, continue using the familiar = syntax without extra { } braces.

from https://herbsutter.com/elements-of-modern-c-style/

> Guideline: Prefer to use initialization with { }, such as vector<int> v = { 1, 2, 3, 4 }; or auto v = vector<int>{ 1, 2, 3, 4 };, because it’s more consistent, more correct, and avoids having to know about old-style pitfalls at all. *In single-argument cases where you prefer to see only the = sign, such as int i = 42; and auto x = anything; omitting the braces is fine.*

from https://herbsutter.com/2013/05/09/gotw-1-solution/

It's hard to defend that `int i{-1};` is more readable than `int i = -1;`. Let's keep simple code simple and use the advanced features everywhere else. [KISS](https://en.wikipedia.org/wiki/KISS_principle)

piraka9011 (approved):
LGTM. Thanks @matthew-reynolds and @bmagyar for taking the time to discuss the approach.

matthew-reynolds:
>It's hard to defend that `int i{-1};` is more readable than `int i = -1;`

I agree with this, but I think the form `int i = {-1};` is just as readable as `int i = -1;`. 

I interpreted Sutter's advice a bit differently from you: We agree that braces are beneficial for non-POD types. Sutter says omitting braces for POD types is fine, in particular for _local_ variables. But for member variables, I think there is still an argument to be made for braces in the form `int i = {-1};`, if only for consistency and the narrowing prevention benefit.

Some examples of what the brace init would look like, in `hardware_interface/test/actuator_state_interface_test.cpp`:

```
  double pos1 = {1.0}, vel1 = {2.0}, eff1 = {3.0};
  double pos2 = {4.0}, vel2 = {5.0}, eff2 = {6.0};
  string name1 = {"name_1"};
  string name2 = {"name_2"};
  ActuatorStateHandle h1 = {name1, &pos1, &vel1, &eff1};
  ActuatorStateHandle h2 = {name2, &pos2, &vel2, &eff2};
```

In `hardware_interface/test/force_torque_sensor_interface_test.cpp`:

```
  double force1[3] = {1.0, 2.0, 3.0};
  double force2[3] = {4.0, 5.0, 6.0};
  double torque1[3] = {-force1[0], -force1[1], -force1[2]};
  double torque2[3] = {-force2[0], -force2[1], -force2[2]};
  string name1 = {"name_1"}, name2 = {"name_2"};
  string frame_id1 = {"frame_1"}, frame_id2 = {"frame_2"};
  ForceTorqueSensorHandle h1 = {name1, frame_id1, force1, torque1};
  ForceTorqueSensorHandle h2 = {name2, frame_id2, nullptr, torque2};
```

IMO these are perfectly readable, and the braces in the primitives make things look more consistent while offering the minor benefit of preventing narrowing.

---

Again, I don't really care what we land on, since the benefit to braces on primitives is so small. Just want to pitch out another viewpoint for consideration before we decide. But I do think we agree that braces should be used for non-POD types - I will update the PR.

bmagyar:
Indeed! Thanks for the discussion guys, as soon as you are done with the last commit I'll go ahead with the merge.

matthew-reynolds:
>as soon as you are done with the last commit I'll go ahead with the merge

By "last commit," do you mean adding braces for non-POD types, or do you mean let's go all-in for braces on all members?

In either case, I will use the `Type i = {-1};` approach over `Type i{-1};` for the familiarity and  readability.

bmagyar:
i = {-1};

matthew-reynolds:
Added braces to member initializers in 1ae77f8

bmagyar (approved):
Thanks very much for this, the future is clean and covered in curly braces ;)

PR #422, opened 2020-01-28
Improve controller and resource filtering for CombinedRobotHW

toliver:
This PR improves the filtering that CombinedRobotHW does on controllers, interfaces and resources before it passes the `start_list` and `stop_list` to the underlying `RobotHW`s.
So now:
- If none of the resources in a certain interface (that had some resources) belong to a certain RobotHW instance, then that interface will be filtered out (for that RobotHW instance).
- If none of the resources in the interfaces of a controller belong to a certain RobotHW instance, then that controller will be filtered out (for that RobotHW instance).

Test cases have been added to cover these two cases.


bmagyar (approved):
Thank you!

PR #423, opened 2020-02-05
Replace boost with std

matthew-reynolds:
Part of #403.

Replaces all usages of boost w/ std equivalents available since C++11 or C++14.

 - `boost::recursive_mutex` -> `std::recursive_mutex`
 - `boost::mutex` -> `std::mutex`
 - `boost::T::scoped_lock` -> `std::lock_guard<T>` (See note 1)
 - `boost::bind` -> `std::bind`
 - `boost::ptr_vector<T>` -> `std::vector<T*>` (See note 2)
 - `boost::lexical_cast<double>` -> `std::stod`

Note 1: `boost::T::scoped_lock` is just a typedef for `boost::unique_lock<T>`. I could have used the corresponding `std::unique_lock<T>`, but for our usage, `std::lock_guard<T>` is fully equivalent and may have slightly less overhead. `std::scoped_lock` supersedes `std::lock_guard`, but is only available in C++17.

Note 2: Although `boost::ptr_vector` offers a lot of functionality/optimizations that `std::vector` doesn't, in our case we're just holding pointers until they are destructed, we never do anything with them, so we don't care about these differences.

### DIFF
--- a/controller_manager/CMakeLists.txt
+++ b/controller_manager/CMakeLists.txt
@@ -10,9 +10,6 @@ find_package(catkin REQUIRED COMPONENTS
   roscpp
 )
 
-# Include Boost.Thread
-find_package(Boost REQUIRED COMPONENTS thread)
-
 catkin_python_setup()
 
 # Declare a catkin package
@@ -27,8 +24,6 @@ catkin_package(
     hardware_interface
     pluginlib
     roscpp
-  DEPENDS
-    Boost
 )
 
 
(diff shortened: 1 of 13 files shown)


matthew-reynolds:
~CI failure looks to be unrelated to these changes. One of the builds passed, one failed during install of `realtime_tools`~

Edit: Thanks to whoever restarted that failed build :)

bmagyar (approved):
Thanks very much for this!
I also agree with both notes, in fact I believe we did the same before in an earlier deboosting works.

PR #426, opened 2020-02-17
fixed startup bug

ThibaultRouillard:
There was a `#KeyError: 'initialized`

Adding the `return` in a `try:` fixes it.

### DIFF
--- a/rqt_controller_manager/src/rqt_controller_manager/controller_manager.py
+++ b/rqt_controller_manager/src/rqt_controller_manager/controller_manager.py
@@ -102,7 +102,9 @@ class ControllerManager(Plugin):
         path = rospack.get_path('rqt_controller_manager')
         self._icons = {'running': QIcon(path + '/resource/led_green.png'),
                        'stopped': QIcon(path + '/resource/led_red.png'),
-                       'uninitialized': QIcon(path + '/resource/led_off.png')}
+                       'uninitialized': QIcon(path + '/resource/led_off.png'),
+                       'initialized': QIcon(path + '/resource/led_off.png') # missing controller statte
+                       }
 
         # Controllers display
         table_view = self._widget.table_view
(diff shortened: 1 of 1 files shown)


rqt_controller_manager/src/rqt_controller_manager/controller_manager.py
```diff
@@ -383,7 +383,11 @@ def data(self, index, role):
 
         if role == Qt.DecorationRole:
             if index.column() == 0:
-                return self._icons[ctrl.state]
+                
+                try:                                
+                    return self._icons[ctrl.state]  #KeyError: 'initialized'
+                except:
```
bmagyar: shouldn't we specifically catch the `KeyError` you put in the comment?

bmagyar (changes requested):

rqt_controller_manager/src/rqt_controller_manager/controller_manager.py
ThibaultRouillard: Actually, there is a missing `ctrl.state`: `initialised`

wxmerkt:
This is currently broken in Melodic from binaries so it would be great to get a fix in ASAP :-).

bmagyar:
@ThibaultRouillard @wxmerkt I've cleaned up the PR and it is now ready for merge. As we don't have CI specifically for rqt plugins I have to ask you guys for confirmation that this fixes the issue you are seeing. As soon as this is in I'm ready to push the next melodic release.

bmagyar (approved):

wxmerkt:
Thank you for the swift action :-) I can confirm this fixes the issue

ThibaultRouillard:
Sorry for the delay. Thank you for accepting the request.

bmagyar:
https://github.com/ros/rosdistro/pull/23860

PR #427, opened 2020-03-11
Bump CMake version to avoid CMP0048

sloretz:
This bumps the minimum CMake version to 3.0.2, which is the [minimum supported by ROS Kinetic](https://www.ros.org/reps/rep-0003.html#kinetic-kame-may-2016-may-2021) and new enough to default to the NEW behavior of [CMP0048](https://cmake.org/cmake/help/v3.0/policy/CMP0048.html). This avoids a CMake warning when building and testing this package in Debian Buster and Ubuntu Focal.

ros/catkin#1052

### DIFF
--- a/combined_robot_hw/CMakeLists.txt
+++ b/combined_robot_hw/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.3)
+cmake_minimum_required(VERSION 3.0.2)
 project(combined_robot_hw)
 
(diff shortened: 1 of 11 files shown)


matthew-reynolds (approved):

bmagyar (approved):
Thank you!

matthew-reynolds:
@bmagyar This could (should?) be backported to Kinetic as well (See [REP 3](https://www.ros.org/reps/rep-0003.html#kinetic-kame-may-2016-may-2021)).

PR #428, opened 2020-03-12
[trivial but useful] Allow spawner to wait indefinitely if requested

v4hn:
`rospy.wait_for_service` does not allow for `0` to wait indefinitely.
Instead `None` has to be passed explicitly.

A user-provided value of 0 leads to an error message and failure at the moment, so it is alright to use the value to indicate indefinite waiting time.

### DIFF
--- a/controller_manager/scripts/spawner
+++ b/controller_manager/scripts/spawner
@@ -80,7 +80,9 @@ def parse_args(args=None):
     parser.add_argument('--namespace', metavar='ns',
                         help='namespace of the controller_manager services')
     parser.add_argument('--timeout', metavar='T', type=float, default=30,
-                        help='how long to wait for controller_manager services when starting up [s] (default: 30)')
+                        help='how long to wait for controller_manager services when starting up [s] (default: 30). <=0 waits indefinitely.')
+    parser.add_argument('--no-timeout', action='store_true',
+                        help='wait for controller_manager services indefinitely (same as `--timeout 0`)')
     parser.add_argument('--shutdown-timeout',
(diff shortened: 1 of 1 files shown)


matthew-reynolds (changes requested):
Seems reasonable, but since this will make the `timeout` behaviour slightly different than `rospy.wait_for_service`, we should add more info. For example, in the help string for the `timeout` argparse argument, maybe add something along the lines of "If <=0, wait indefinitely".

Alternatively/in addition, we could add a `--no-timeout` argument. Might be clearer to users. Thoughts?

destogl:
+1 for --no-timeout

bmagyar:
I see value in both using the number to encode wait time (if passed in through a parameter) and the more verbose option. Can we have both please? ;)

v4hn:
Wow, it *is* a trivial patch obviously improving applicability in a case that was not documented *at all* before.
I barely care whether it gets merged or not, I just took a minute to pull-request what I typed locally.

So I'm a bit surprised to get further request for icing and sprinkles. *sigh*
Anyway addressed.

matthew-reynolds (approved):
Looks great, thanks so much for iterating.

> It is a trivial patch obviously improving applicability in a case that was not documented at all before

Indeed, thanks for pointing out the flaw. Once it was identified, I figured we might as well handle it fully. Always good to strive for the best code we can :)

bmagyar:
@v4hn indeed sometimes it is unusual to be on the receiving end of reviews ;)

bmagyar (approved):
Thanks very much for pointing this out and providing a fix!

PR #430, opened 2020-03-15
Add function specifiers and modernize constructors

matthew-reynolds:
Part of #403.

Add function specifiers (`override`, `noexcept`) and cleanup ctors & dtors (`= default`, `= delete`, and removing unnecessary).

A couple of notes:
 - Per [CppCoreGuidelines C.128](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c128-virtual-functions-should-specify-exactly-one-of-virtual-override-or-final), **Virtual functions should specify exactly one of virtual, override, or final**. Therefore, lots of `virtual` specifiers were removed in favour of `override`s.
 - We had a lot of unnecessary empty constructors and overriding destructors. Wherever possible, these have been removed.
 - `throw()` -> `noexcept`. I did not add any additional `noexcept`s though, only updated the existing ones in the custom exceptions.
 - `ImuSensorHandle::Data::Data()` must be user-defined, it cannot be `= default`. See https://stackoverflow.com/a/17436088/1932358

---

This was mostly done using `clang-tidy`:
```
Checks: '-*,
        modernize-use-equals-default,
        modernize-use-equals-delete,
        modernize-use-override,
        modernize-use-noexcept
        '
```

### DIFF
--- a/combined_robot_hw/include/combined_robot_hw/combined_robot_hw.h
+++ b/combined_robot_hw/include/combined_robot_hw/combined_robot_hw.h
@@ -52,7 +52,7 @@ namespace combined_robot_hw
 class CombinedRobotHW : public hardware_interface::RobotHW
 {
 public:
-  virtual ~CombinedRobotHW(){}
+  ~CombinedRobotHW() override = default;
 
   /** \brief The init function is called to initialize the RobotHW from a
    * non-realtime thread.
@@ -64,7 +64,7 @@ public:
    *
    * \returns True if initialization was successful
    */
-  virtual bool init(ros::NodeHandle& root_nh, ros::NodeHandle &robot_hw_nh);
+  bool init(ros::NodeHandle& root_nh, ros::NodeHandle &robot_hw_nh) override;
 
 
   /**
@@ -72,15 +72,15 @@ public:
    * with regard to necessary hardware interface switches and prepare the switching. Start and stop list are disjoint.
    * This handles the check and preparation, the actual switch is commited in doSwitch()
    */
-  virtual bool prepareSwitch(const std::list<hardware_interface::ControllerInfo>& start_list,
-                             const std::list<hardware_interface::ControllerInfo>& stop_list);
+  bool prepareSwitch(const std::list<hardware_interface::ControllerInfo>& start_list,
+                     const std::list<hardware_interface::ControllerInfo>& stop_list) override;
 
   /**
    * Perform (in realtime) all necessary hardware interface switches in order to start and stop the given controllers.
    * Start and stop list are disjoint. The feasability was checked in prepareSwitch() beforehand.
    */
-  virtual void doSwitch(const std::list<hardware_interface::ControllerInfo>& start_list,
-                        const std::list<hardware_interface::ControllerInfo>& stop_list);
+  void doSwitch(const std::list<hardware_interface::ControllerInfo>& start_list,
+                const std::list<hardware_interface::ControllerInfo>& stop_list) override;
 
   /**
    * Reads data from the robot HW
@@ -88,7 +88,7 @@ public:
    * \param time The current time
    * \param period The time passed since the last call to \ref read
    */
-  virtual void read(const ros::Time& time, const ros::Duration& period);
+  void read(const ros::Time& time, const ros::Duration& period) override;
 
   /**
    * Writes data to the robot HW
@@ -96,7 +96,7 @@ public:
    * \param time The current time
    * \param period The time passed since the last call to \ref write
    */
-  virtual void write(const ros::Time& time, const ros::Duration& period);
+  void write(const ros::Time& time, const ros::Duration& period) override;
 
 protected:
   ros::NodeHandle root_nh_;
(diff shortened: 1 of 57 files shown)


combined_robot_hw/include/combined_robot_hw/combined_robot_hw.h
```diff
@@ -52,7 +52,6 @@ namespace combined_robot_hw
 class CombinedRobotHW : public hardware_interface::RobotHW
 {
 public:
-  virtual ~CombinedRobotHW(){}
```
bmagyar, line 55: probably still needed but could be ` virtual ~CombinedRobotHW() = default;`

hardware_interface/include/hardware_interface/imu_sensor_interface.h
```diff
@@ -46,6 +46,7 @@ class ImuSensorHandle
 public:
   struct Data
   {
+    // Note: User-provided constructor required due to a defect in the standard. See https://stackoverflow.com/a/17436088/1932358
     Data() {};
```
bmagyar:
```suggestion
    Data() {}
```

bmagyar (commented):
In general it looks very good to me, love to see code improvements with a negative balance sheet ;) 
I am unsure about the virtual destructors though...
* A tiny bit unsure whether without declaring any, not even an override, derived class destructors will be virtual too
* Whether we should keep the virtual desctructor as a reminder for people using this class to define one. Yes this class gets it from it's parent but this class is meant to be subclassed from so I think we should not rely on implicit virtuals.

matthew-reynolds:
> A tiny bit unsure whether without declaring any, not even an override, derived class destructors will be virtual too

Yep, the derived class destructor will be virtual if there's a virtual destructor anywhere in the inheritance chain. As a quick sanity check: http://cpp.sh/4d4l

> Whether we should keep the virtual destructor as a reminder for people using this class to define one. Yes this class gets it from it's parent but this class is meant to be subclassed from so I think we should not rely on implicit virtuals.

Hmm, I maybe see the argument for keeping it to make it extra clear that the class has a virtual destructor, but I don't think it's really important to tell that to the author of subclasses. They don't even need to be aware of the virtual dtor, since it doesn't affect how they write their code - If they need a destructor to do some cleanup for stuff in their class, they can add one, but if they don't need one, they don't need to add one. I think it's unnecessary to "remind people to define one" since not all subclasses will require a user-defined destructor, and they can just follow the regular rules about when to define one. 

I also think it's still pretty clear that `CombinedRobotHW` has a virtual destructor, since `RobotHW` has one. Similarly, that `Controller` has a virtual destructor, since `ControllerBase` has one.

hardware_interface/include/hardware_interface/imu_sensor_interface.h
matthew-reynolds: Good catch!

matthew-reynolds: Fixed in 70231fee

bmagyar:
It wasn't hard to convince me ;)  Keeping `virtual` as self-documentation to people writing derived classes is already somewhat disrupted by `override` where using `virtual` is superfluous.

jordan-palacios:
Out of curiosity what's the criteria for `default`ing constructors instead of removing them?

For example, as far as I can tell, [this](https://github.com/ros-controls/ros_control/pull/430/files#diff-007efb4195f9435f1f8f9797dd910d5dR43) one could be removed too.

matthew-reynolds:
Which one is that link pointing to?

The rationale was to remove as many as possible, and then only `default` ones that have other overloads. If there are any left that can be removed, they're probably just a mistake or oversight. Happy to update the PR.

(I would also like all those ctor overloads to be `explicit` too, but that will have to target Noetic since it's breaking. See [CppCoreGuidelines C.46](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c46-by-default-declare-single-argument-constructors-explicit) and https://abseil.io/tips/142. That's on my list of follow-ups)

hardware_interface/include/hardware_interface/force_torque_sensor_interface.h
```diff
@@ -40,7 +40,7 @@ namespace hardware_interface
 class ForceTorqueSensorHandle
 {
 public:
-  ForceTorqueSensorHandle() {}
+  ForceTorqueSensorHandle() = default;
```
jordan-palacios, line 44: I was referring to this one.

matthew-reynolds: All the `XYZHandle`s have a default constructor and parameterized constructor, so I left the default actors (Updated to `= default`) to avoid breaking API/ABI.

jordan-palacios: I see now. Thanks!

bmagyar (approved):

bmagyar:
Rebased for latest melodic, ready to merge once CI is happy

Issue #431, opened 2020-03-20
spawner script does not wait for (sim) clock before waiting for services


guillaumeautran:
With a simulation, the Gazebo clock could potentially not be available right away (takes time to be published or started unpublished). The spawner script proceeds to loading the controllers and eventually times out if the clock does not arrive in time. 

A change to have the spawner script wait for the clock to be published would make the loading of controller more resilient in simulation when the simulated clock could be delayed.

PR #432, opened 2020-03-20
SPAWNER: Wait for the clock to be available before proceeding

guillaumeautran:
This patch is to resolve an issue when the simulated clock is not published right away (coming from a remote gazebo instance for example).
If the clock is not published right away, the rosservice wait_for_service times out prematurely and the controllers fail to be loaded since the system is not yet ready (no Gazebo clock).

With no, gazebo system, the clock will be ready immediately and the spawner will proceed.

Issue: #431

### DIFF
--- a/controller_manager/scripts/spawner
+++ b/controller_manager/scripts/spawner
@@ -116,6 +116,13 @@ def main():
     unload_controller_service = robot_namespace+"controller_manager/unload_controller"
     switch_controller_service = robot_namespace+"controller_manager/switch_controller"
 
+    # Wait for the clock to be published
+    if rospy.get_param('/use_sim_time', False):
+        while rospy.get_rostime() == rospy.Time(0):
+            rospy.loginfo_throttle(30, "Waiting for /clock to be available...")
+            rospy.sleep(0.2)
+        rospy.loginfo("/clock is published. Proceeding to load the controller(s).")
+
     try:
         # loader
         rospy.loginfo("Controller Spawner: Waiting for service "+load_controller_service)


matthew-reynolds:
It looks like the same issue was fixed upstream for C++ in `roscpp` [here](https://github.com/ros/ros_comm/pull/1638), would this be better suited as an upstream fix for Python in `rospy`?

guillaumeautran:
I can make a similar fix in rospy if you think it is "safe" and has no side-effects.

guillaumeautran:
Actually, @matthew-reynolds , the change that you pointed is actually the opposite to what this change is suggesting. In the referenced change, the service wait fails because it uses the ros::Duration as opposed to the ros::WallDuration. The former takes in consideration the `/clock` topic while the later does not. 
rospy wait_for_service does not currently rely on the simulated clock, which is consistent with the roscpp version. So, no, modifying ros_comm is not the right approach.

controller_manager/scripts/spawner
```diff
@@ -116,6 +116,12 @@ def main():
     unload_controller_service = robot_namespace+"controller_manager/unload_controller"
     switch_controller_service = robot_namespace+"controller_manager/switch_controller"
 
+    # Wait for the clock to be published
+    while rospy.get_rostime() == rospy.Time(0):
+        rospy.loginfo_throttle(30, "Waiting for the ROS Clock to be available...")
+        rospy.sleep(0.2)
+    rospy.loginfo("ROS Clock acquired. Proceeding to load the controller(s).")
+
```
matthew-reynolds: Maybe let's wrap this in a check of the `/use_sim_time` param, and then update "ROS Clock" to "/clock" or "Simulated clock" or something along those lines?

I think that would be clearer, and it would prevent a confusing log message about "acquiring the clock" on wall-time systems.

matthew-reynolds (changes requested):
I see, I misunderstood the problem. I thought by "`wait_for_service` times out prematurely" you meant that there was a problem with `wait_for_service`. The issue is really that, in our use-case, the timeout should not start counting down until the system is "ready", which we're measuring by seeing whether Gazebo's clock is running.

That makes sense to me, seems like a reasonable change. Since `rospy.get_rostime()` should only return 0 when the `/use_sim_time` param is true, can we leverage that to provide more precise log messages?

matthew-reynolds:
@guillaumeautran @mikepurvis Bump. Is this approach being used at Clearpath? Assuming it's seen some use and has had a positive impact, we should try to get it tidied up and merged.

Edit: And re-targetted to Noetic

bmagyar:
I think this is a useful feature we could incorporate, happy to cherry-pick to melodic too but indeed let's aim the PR to noetic first. 

If @matthew-reynolds is happy I'm happy to merge & release to Noetic

guillaumeautran:
Very sorry for dropping the ball on this one. I know its been a while..... 
The patch is now rebased to target Noetic (instead of Melodic). 

@matthew-reynolds WRT wrapping the check in the `use_sim_time` param, any clean suggestions on implementation? The sim time check is not used anywhere in `ros-control` (from what I can tell). Does not mean it should not be done but just making sure that's what we want before introducing a new dependency here.  

Here is what I am thinking:
```
    # Wait for the clock to be published
    if rospy.get_param('/use_sim_time', False):
        while rospy.get_rostime() == rospy.Time(0):
            rospy.loginfo_throttle(30, "Waiting for /clock to be available...")
            rospy.sleep(0.2)
        rospy.loginfo("/clock is published. Proceeding to load the controller(s).")
```

matthew-reynolds:
> The sim time check is not used anywhere in `ros-control` (from what I can tell). Does not mean it should not be done but just making sure that's what we want before introducing a new dependency here.

I think that's fine. In general I agree, `ros-control` should be agnostic to time base, but I think it's appropriate here in a spawner script.

> `if rospy.get_param('/use_sim_time', False):`

Shouldn't this be the other way (`True`)? We only want to do this check if using sim time (e.g. from Gazebo), right?

guillaumeautran:
The statement:
`if rospy.get_param('/use_sim_time', False):`
Means that if the parameter is not defined, the returned value (default) should be `False` hence bypassing the check altogether. I am assuming that the rosparam will only be defined if in simulation...

matthew-reynolds (approved):
> if the parameter is not defined, the default should be `False`

doh. Been too long since I wrote Python ROS.

LGTM, thanks :) We'll let Bence take a look and merge if he's happy.

guillaumeautran:
Perfect. Thanks much!

guillaumeautran:
ping @bmagyar for review.

bmagyar (approved):

PR #433, opened 2020-03-26
[hardware_interface::RobotHW] Extend documentation of read and write methods

fjp:
This PR is related to [ros-controls/ros_controllers/#474](https://github.com/ros-controls/ros_controllers/pull/474) and updates/extends the documentation for `read` and `write` methods.

The PR adds a `\note` to the documentation regarding the directional perspective of these two methods. 

I've also added a [named member group](http://www.doxygen.nl/manual/grouping.html#memgroup) "Control Loop". Previously, `read` and `write` were grouped under "Hardware Interface Switching", which is not correct  I guess.

### DIFF
--- a/hardware_interface/README.md
+++ b/hardware_interface/README.md
@@ -1,0 +1,299 @@
+# Hardware Interfaces           {#mainpage}
+
+hardware interfaces are used by ROS control in conjunction with one of the 
+available ROS controllers to send (\ref hardware_interface::RobotHW::write) 
+commands to the hardware and receive (\ref hardware_interface::RobotHW::read) 
+states from the robot's resources (joints, sensors, actuators).
+
+A list of available hardware interfaces (provided via the HardwareResourceManager) 
+as of this writing:
+
+- [JointCommandInterface](\ref hardware_interface::JointCommandInterface): 
+  hardware interface to support commanding and reading the state of an array of
+  joints. Note that these commands can have any semantic meaning as long as each
+  can be represented by a single double, they are not necessarily effort commands. 
+  To specify a meaning to this command, see the derived classes:
+  - [EffortJointInterface](\ref hardware_interface::EffortJointInterface): 
+    for commanding and reading effort-based joints.
+  - [VelocityJointInterface](\ref hardware_interface::VelocityJointInterface): 
+    for commanding and reading velocity-based joints.
+  - [PositionJointInterface](\ref hardware_interface::PositionJointInterface): 
+    for commanding and reading position-based joints.
+- [JointStateInterfaces](\ref hardware_interface::JointStateInterface): 
+  hardware interface to support reading the state of an array of named joints, 
+  each of which has some position, velocity, and effort (force or torque).
+- [ActuatorStateInterfaces](\ref hardware_interface::ActuatorStateInterface): 
+  hardware interface to support reading the state of an array of named actuators,
+  each of which has some position, velocity, and effort (force or torque).
+- [ActuatorCommandInterfaces](\ref hardware_interface::ActuatorCommandInterface)
+  - [EffortActuatorInterface](\ref hardware_interface::EffortActuatorInterface)
+  - [VelocityActuatorInterface](\ref hardware_interface::VelocityActuatorInterface)
+  - [PositionActuatorInterface](\ref hardware_interface::PositionActuatorInterface)
+- Force-torque sensor Interface
+- IMU sensor Interface
+
+Note that \ref hardware_interface::JointCommandInterface allows both reading 
+joint state and commanding [effort|velocity|position]-based joints 
+(see this [answer](https://answers.ros.org/question/209619/differences-between-hardware-interfaces/?answer=209636#post-id-209636)).
+
+## Setting up a new robot
+
+The following example explains how to set up a new robot to work with the 
+[controller_manager](http://wiki.ros.org/controller_manager). When you make your 
+robot support one or more of the standard interfaces, you will be able to take 
+advantage of a large library of controllers (see 
+[ros_controllers](https://github.com/ros-controls/ros_controllers)) 
+that work on the standard interfaces mentioned above.
+
+![hw_interface_2.png](https://raw.githubusercontent.com/wiki/ros-controls/ros_control/hw_interface_2.png)
+
+Assuming a robot with two joints: A & B where both joints are position controlled. 
+Such a robot is defined with class derived from \ref hardware_interface::RobotHW
+that should provide the standard [PositionJointInterface](\ref hardware_interface::PositionJointInterface) and the 
+[JointStateInterface](\ref hardware_interface::JointStateInterface), so it can re-use all controllers that are already 
+written to work with the [PositionJointInterface](\ref hardware_interface::PositionJointInterface) and the 
+[JointStateInterface](\ref hardware_interface::JointStateInterface). The code would look like this:
+
+
+```cpp
+#include <hardware_interface/joint_command_interface.h>
+#include <hardware_interface/joint_state_interface.h>
+#include <hardware_interface/robot_hw.h>
+
+class MyRobot : public hardware_interface::RobotHW
+{
+public:
+  MyRobot() 
+  { 
+    // Initialization of the robot's resources (joints, sensors, actuators) and
+    // interfacs can be done here or inside init().
+  }
+
+  bool init(ros::NodeHandle &root_nh, ros::NodeHandle &robot_hw_nh)
+  {
+    // Create a JointStateHandle for each joint and register them with the 
+    // JointStateInterface.
+    hardware_interface::JointStateHandle state_handle_a("A", &pos[0], &vel[0], &eff[0]);
+    jnt_state_interface.registerHandle(state_handle_a);
+
+    hardware_interface::JointStateHandle state_handle_b("B", &pos[1], &vel[1], &eff[1]);
+    jnt_state_interface.registerHandle(state_handle_b);
+
+    // Register the JointStateInterface containing the read only joints
+    // with this robot's hardware_interface::RobotHW.
+    registerInterface(&jnt_state_interface);
+
+    // Create a JointHandle (read and write) for each controllable joint
+    // using the read-only joint handles within the JointStateInterface and 
+    // register them with the JointPositionInterface.
+    hardware_interface::JointHandle pos_handle_a(jnt_state_interface.getHandle("A"), &cmd[0]);
+    jnt_pos_interface.registerHandle(pos_handle_a);
+
+    hardware_interface::JointHandle pos_handle_b(jnt_state_interface.getHandle("B"), &cmd[1]);
+    jnt_pos_interface.registerHandle(pos_handle_b);
+
+    // Register the JointPositionInterface containing the read/write joints
+    // with this robot's hardware_interface::RobotHW.
+    registerInterface(&jnt_pos_interface);
+
+    return true;
+  }
+
+private:
+  hardware_interface::JointStateInterface jnt_state_interface;
+  hardware_interface::PositionJointInterface jnt_pos_interface;
+
+  // Data member array to store the controller commands which are sent to the 
+  // robot's resources (joints, actuators)
+  double cmd[2];
+
+  // Data member arrays to store the state of the robot's resources (joints, sensors)
+  double pos[2];
+  double vel[2];
+  double eff[2];
+};
+```
+
+This code represents a custom robot and is required to control it. 
+The functions above are designed to give the controller manager (and the 
+controllers inside the controller manager) access to the joint state of a
+custom robot, and to command it. When the controller manager runs, the 
+controllers will read from the pos, vel and eff variables of the custom robot 
+hardware interface, and the controller will write the desired command into the 
+cmd variable. It's mandatory to make sure the pos, vel and eff variables always 
+have the latest joint state available, and to make sure that whatever is written
+into the cmd variable gets executed by the robot. This can be done by implementing 
+\ref hardware_interface::RobotHW::read() and a \ref hardware_interface::RobotHW::write()
+methods. A node's `main()` function can be implemented like this:
+
+```cpp
+#include <ros/ros.h>
+#include <my_robot/my_robot.h>
+#include <controller_manager/controller_manager.h>
+
+int main(int argc, char **argv)
+{
+    // Initialize the ROS node
+    ros::init(argc, argv, "my_robot");
+    
+    // Create an instance of your robot so that this instance knows about all 
+    // the resources that are available.
+    MyRobot::MyRobot robot;
+
+    // Create an instance of the controller manager and pass it the robot, 
+    // so that it can handle its resources.
+    controller_manager::ControllerManager cm(&robot);
+    
+    // Setup a separate thread that will be used to service ROS callbacks.
+    ros:AsyncSpinner spinner(1);
+    spinner.start();
+    
+    // Setup for the control loop.
+    ros::Time prev_time = ros::Time::now();
+    ros::Rate rate(10.0); // 10 Hz rate
+    
+    while (ros::ok())
+    {
+        // Basic bookkeeping to get the system time in order to compute the control period.
+        const ros::Time     time   = ros:Time::now();
+        const ros::Duration period = time - prev_time;
+        
+        // Execution of the actual control loop.
+        robot.read();
+        cm.update(time, period);
+        root.write();
+        
+        // All these steps keep getting repeated with the specified rate.
+        rate.sleep();
+    }
+    return 0;
+}
+```
+
+As the image above suggests, a custom robot hardware interface is not 
+limited to be composed of only one single interface. The robot can provide as 
+many interfaces as required. It could for example provide both the 
+\ref hardware_interface::PositionJointInterface and the 
+\ref hardware_interface::VelocityJointInterface, and many more.
+
+
+## Resource Management
+
+The controller manager keeps track of which resources are in use by each of the 
+controllers. A resource can be something like 'right_elbow_joint', 'base', 
+'left_arm', 'wrist_joints'. Pretty much anything a specific robot can consist of. 
+Put simply, resources are specified in the robot's hardware interface (derived from 
+\ref hardware_interface::RobotHW). And a robot is represented by a bunch of 
+hardware interfaces, where one is a set of similar resources. For example, the 
+\ref hardware_interface::PositionJointInterface groups the position controlled 
+joints as resources. It is also possible to implement a custom hardware interface, 
+and define custom resources. When controllers are getting initialized, they 
+request a number of resources from the hardware interface; these requests get 
+recorded by the controller manager. So the controller manager knows exactly 
+which controller has requested which resources.
+
+The RobotHW class has a simple default implementation for resource management: 
+it simply prevents two controllers that are using the same resource to be 
+running at the same time. Note that two controllers that are using the same 
+resource can be loaded at the same time, but can't be running at the same time. 
+If this simple resource management scheme fits to a robot, nothing else needs to
+be done, the controller manager will automatically apply this scheme. If a robot 
+needs a different scheme, it is possible to create a custom one, by implementing 
+one single function:
+
+```cpp
+class MyRobot : public hardware_interface::RobotHW
+{
+public:
+  MyRobot() 
+  {
+    // register hardware interfaces interfaces
+    ...
+    ... (see the code above)
+    ...
+
+
+    // Implement robot-specific resouce management
+    bool checkForConflict(const std::list<ControllerInfo>& info) const
+    {
+       // This list of controllers cannot be running at the same time
+       ... 
+       return true;
+
+       // This list of controller can be running at the same time
+       ...
+       return false;
+    }
+  }
+};
+```
+
+The input to the \ref hardware_interface::RobotHW::checkForConflict method is a list of 
+controller info objects. Each of these objects matches to one single controller, 
+and contains all the info about that controller. This info includes the 
+controller name, controller type, hardware interface type, and the list of 
+resources that are claimed by the controller. Based on all this info, it is
+possible to come up with a custom scheme to decide if the given list of 
+controllers is allowed to be running at the same time.
+
+
+## Creating a robot-specific interface
+
+The standard interfaces are helpful to avoid writing a whole new set of 
+controllers for a robot, and to take advantage of the libraries of existing 
+controllers. But in case a robot has some features that are not supported by the 
+standard interfaces it is possible to leverage the standard interfaces 
+(and re-use the standard controllers) for the features of a robot that are standard. 
+And at the same time expose robot-specific features in a robot-specific interface. 
+The image shown above shows a robot with both standard and robot-specific interfaces.
+
+The code for such a robot looks like this:
+
+```cpp
+class MyRobot : public hardware_interface::RobotHW
+{
+public:
+  MyRobot() 
+  {
+    // Register the joint state and position interfaces.
+    ...
+    ... (see the code above)
+    ...
+
+    // Register some robot specific interfaces.
+    registerInterface(&cool_interface);
+  }
+
+  private:
+    MyCustomInterface cool_interface;
+};
+```
+
+Anotehr way is to register the MyRobot class itself:
+
+```cpp
+class MyRobot : public hardware_interface::RobotHW, public hardware_interface::HardwareInterface
+{
+public:
+  MyRobot() 
+  {
+    // Register the joint state and position interfaces.
+    ...
+    ... (see the code above)
+    ...
+
+    // Register the MyRobot class itself to make the 'someCoolFunction' available.
+    // The MyRobot class inherits from HardwareInterface to make this possible.
+    registerInterface(this);
+  }
+
+  void someCoolFunction();
+ 
+};
+```
+
+With this, the custom interfaces could be nothing more than adding any number of 
+function calls to a custom robot class, and registering the robot class itself. 
+These robot-specific functions will only be available to controllers that are 
+specifically designed for the custom robot, but at the same time, the robot will 
+still work with standard controllers using the standard interfaces of the robot.
--- a/hardware_interface/include/hardware_interface/robot_hw.h
+++ b/hardware_interface/include/hardware_interface/robot_hw.h
@@ -208,10 +208,10 @@ public:
 
   /** \brief Writes commands from controllers to the robot hardware.
    * 
-   * The write method is part of the control loop cycle (read, update, write) 
+   * The write method is part of the control loop cycle (\ref read, update, \ref write) 
    * and used to send out commands, via the hardware interfaces 
-   * (eg. JointCommandInterface and its subclasses), to your robot's hardware 
-   * esources (joints, actuators). Inside a node that handles the control loop, 
+   * (eg. JointCommandInterface and its subclasses), to the robot's hardware 
+   * resources (joints, actuators). Inside a node that handles the control loop, 
    * this method should be called after \ref read and controller_manager::ControllerManager::update.
    * 
    * \note The name \ref write refers to writing (sending out) data commands,
--- a/hardware_interface/include/hardware_interface/robot_hw.h
+++ b/hardware_interface/include/hardware_interface/robot_hw.h
@@ -51,8 +51,18 @@ namespace hardware_interface
  * hardware.
  *
  * The hardware interface map (\ref interfaces_) is a 1-to-1 map between
- * the names of interface types derived from \ref HardwareInterface  and
+ * the names of interface types derived from \ref HardwareInterface and
  * instances of those interface types.
+ * 
+ * The registration (\ref registerInterface) of interfaces can be done in the 
+ * constructor or init() of a custom robot hardware class.
+ * 
+ * The robot state is described by data members of a class derived from this 
+ * base interface. The names of these data members should give semantic meaning 
+ * (pos, vel, eff) to the registered resources (joints, sensors, actuators). 
+ * Commands from controllers are represented by a member array (e.g. cmd)
+ * and populated by the controller's update method, 
+ * (controller_interface::ControllerBase::update()).
  *
  */
 class RobotHW : public InterfaceManager
@@ -70,6 +80,13 @@ public:
 
   /** \brief The init function is called to initialize the RobotHW from a
    * non-realtime thread.
+   * 
+   * The registration of data members to interfaces is done via 
+   * \ref JointStateInterface, for read-only joints, or \ref JointCommandInterface, 
+   * for joints that accept commands and provide feedback (read and write), 
+   * including its derived classes (e.g. \ref VelocityJointInterface).
+   * These interfaces make use of \ref JointHandle or \ref JointStateHandle through 
+   * the \ref HardwareResourceManager.
    *
    * \param root_nh A NodeHandle in the root of the caller namespace.
    *
@@ -129,17 +146,16 @@ public:
   /** \name Hardware Interface Switching
    *\{*/
 
-  /** \brief Check (in non-realtime) if given controllers could be started and stopped from the current state of the RobotHW
-   * with regard to necessary hardware interface switches and prepare the switching.
-   * 
-   * Start and stop list are disjoint.
+  /**
+   * Check (in non-realtime) if given controllers could be started and stopped from the current state of the RobotHW
+   * with regard to necessary hardware interface switches and prepare the switching. Start and stop list are disjoint.
    * This handles the check and preparation, the actual switch is commited in doSwitch()
    */
   virtual bool prepareSwitch(const std::list<ControllerInfo>& /*start_list*/,
                              const std::list<ControllerInfo>& /*stop_list*/) { return true; }
 
-  /** \brief Perform (in realtime) all necessary hardware interface switches in order to start and stop the given controllers.
-   * 
+  /**
+   * Perform (in realtime) all necessary hardware interface switches in order to start and stop the given controllers.
    * Start and stop list are disjoint. The feasability was checked in prepareSwitch() beforehand.
    */
   virtual void doSwitch(const std::list<ControllerInfo>& /*start_list*/,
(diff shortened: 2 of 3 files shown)


fjp:
Using [`rosdoc_lite`](http://wiki.ros.org/rosdoc_lite) locally seems to work except that I don't know why "Hardware Interface Switching" is grouped differently than "Control Loop" and "Resource Management":

image

hardware_interface/include/hardware_interface/robot_hw.h
```diff
@@ -148,33 +152,72 @@ class RobotHW : public InterfaceManager
...
+
+  
+  /** \name Control Loop
+   * 
+   */
+   /**\{*/
 
-  /**
-   * Reads data from the robot HW
+  /** \brief Used to read data from the robot hardware (represented by its interfaces) and populate the raw data.
```
matthew-reynolds: Let's keep this shorter and in the same tense as the rest of the documentation. `Read data from the robot hardware`, for example.

hardware_interface/include/hardware_interface/robot_hw.h
```diff
@@ -148,33 +152,72 @@ class RobotHW : public InterfaceManager
...
+   *
+   * The read method is part of the control loop cycle (read, update, write) and used to populate the raw data with
```
matthew-reynolds, line 176:
```suggestion
   * The read method is part of the control loop cycle (read, update, write) and is used to populate the raw data with 
```

hardware_interface/include/hardware_interface/robot_hw.h
```diff
@@ -148,33 +152,72 @@ class RobotHW : public InterfaceManager
...
+   * 
+   */
+   /**\{*/
 
-  /**
-   * Reads data from the robot HW
+  /** \brief Used to read data from the robot hardware (represented by its interfaces) and populate the raw data.
+   *
+   * The read method is part of the control loop cycle (read, update, write) and used to populate the raw data with 
+   * the state of your robot's hardware resources (joints, sensors, actuators). In your node that handles the control loop,
```
matthew-reynolds, line 177:
```suggestion
   * the state of the robot's hardware resources (joints, sensors, actuators).
```

hardware_interface/include/hardware_interface/robot_hw.h
```diff
@@ -148,33 +152,72 @@ class RobotHW : public InterfaceManager
...
+   * you should call this method before calling controller_manager::ControllerManager::update and write().
```
matthew-reynolds: I think this can be simplified, and also let's remove "your".

```suggestion
   * The read method is part of the control loop cycle (read, update, write) and is used to populate the robot state from
   * the robot's hardware resources (joints, sensors, actuators). This method should be called before controller_manager::ControllerManager::update() and write().
```

hardware_interface/include/hardware_interface/robot_hw.h
```diff
@@ -148,33 +152,72 @@ class RobotHW : public InterfaceManager
...
+   * 
+   * The raw data describes regions in memory of your custom robot hardware interface (subclassed from hardware_interface::RobotHW),
+   * which is usually represented by member arrays having names that give semantic meaning (pos, vel, eff) to the registered resources.
+   * The registration of memory regions (assigning pointers) is done via JointStateInterface, for read-only joints,
+   * or JointCommandInterface, for joints that accept commands and provide feedback (read and write), and its subclasses
+   * (eg. VelocityJointInterface) that make use of JointHandle or JointStateHandle through the HardwareResourceManager.
```
matthew-reynolds: This seems a little out of place to me. Registering interfaces should probably be up in the documentation for `init()`, since that's where it normally happens. I don't think we need discussion about how to store the robot state in `RobotHW`, since that's pretty trivial and also very much up to the implementation.

hardware_interface/include/hardware_interface/robot_hw.h
```diff
@@ -148,33 +152,72 @@ class RobotHW : public InterfaceManager
...
+   * 
+   * \note The name read, of this method, depends on one's directional perspective. 
+   * In the context of the robot hardware_interface::RobotHW it refers to reading state from the robot hardware into the raw data.
+   * This is opposite to the directional perspective where write() is used to fill (write to) 
+   * the raw data with the robot's resource state. Note that this is not the intended approach.
+   *
```
matthew-reynolds: I don't think we need this at all, I think your additions above explain this well enough, but if you think we need the additional clarity let's simplify it to make it even easier to understand:

```suggestion
   * \note The name 'read()' refers to reading state from the hardware.
   * This complements 'write()', which refers to writing commands to the hardware.
   * 
```

matthew-reynolds (changes requested):
IMO the `\brief` on `prepareSwitch` and `doSwitch` doesn't add anything, and is too long to be a brief. I think it's fine without these changes, but if you want to give it a brief, then a shorter, less detailed summary might be better than just pulling the first sentence. Something like `Check (in non-realtime) if the given controllers can be started and stopped.`, and then leave the rest of the body as it was before to add the necessary details.

I added some suggestions below, everything I wrote for `read()` can be translated to `write()` as well. The suggestions probably violate the line length limits, please ignore my bad formatting :)

Overall I think it's great but the language can be simplified a lot. In particular, I'd like to stay away from getting too close to implementation details, since we shouldn't try to force a particular style of implementation. I'm also not a fan of "raw data", so in my suggestions I used "robot state" and "command" instead.

fjp:
Thanks a lot for this thorough review. I'll soon take care of your suggestions.

hardware_interface/include/hardware_interface/robot_hw.h
fjp: I'd prefer to keep it the way you suggested and don't remove it completely. Just for extra clarity.

hardware_interface/include/hardware_interface/robot_hw.h
fjp: 👍

hardware_interface/include/hardware_interface/robot_hw.h
fjp: 👍

hardware_interface/include/hardware_interface/robot_hw.h
fjp: You are right, this is probably not the best place. I've put the description about interface registration here because I wasn't aware that this is usually done inside `init()`. According to the [example in the wiki](https://github.com/ros-controls/ros_control/wiki/hardware_interface#setting-up-a-new-robot) the registration is done in the constructor and I've seen other examples doing this as well, probably because they follow the wiki example. Looking at [ros_control_boilerplate](https://github.com/PickNikRobotics/ros_control_boilerplate/blob/5e40c7e875f7ed58ed061f37d4898865e382170e/src/generic_hw_interface.cpp#L66) it's done in an `init()` method. Could you please clarify where the registration of interfaces should take place or is it just personal preference?

In case it's done inside `init()`, I'll move this part of the documentation there after adjusting it accordingly.

Do you know why the `init()` method of [ros_control_boilerplate](https://github.com/PickNikRobotics/ros_control_boilerplate/blob/5e40c7e875f7ed58ed061f37d4898865e382170e/src/generic_hw_interface.cpp#L66) isn't using the same parameters as the [base class](https://github.com/ros-controls/ros_control/blob/b78bccb5f3768d1726db79acbe8837cc5187b515/hardware_interface/include/hardware_interface/robot_hw.h#L81). So far I haven't seen anyone overriding and calling [`init(ros::NodeHandle& /*root_nh*/, ros::NodeHandle &/*robot_hw_nh*/)`](http://docs.ros.org/melodic/api/hardware_interface/html/c++/classhardware__interface_1_1RobotHW.html#a39c9ede49f3219ae2f668317c55022b6) except [this test](https://github.com/ros-controls/ros_control/blob/melodic-devel/combined_robot_hw_tests/test/combined_robot_hw_test.cpp#L38). I think, if the intended place to register interfaces is in `init(ros::NodeHandle& , ros::NodeHandle &)` and not the constructor of RobotHW we should document it in the docstring of `init()` and update the wiki example. Please let me know how you think about this.

The following may be a bit off-topic regarding this review but it is related:
As far as I know `init(ros::NodeHandle& , ros::NodeHandle &)` (or just `init()` w/o parameters) - when used for registering interfaces - is called from a node's `main()` function. Wouldn't it be possible to move the call of `init(ros::NodeHandle& , ros::NodeHandle &)` inside the [controller_manager constructor](https://github.com/ros-controls/ros_control/blob/b78bccb5f3768d1726db79acbe8837cc5187b515/controller_manager/src/controller_manager.cpp#L45) to avoid having the user calling `init()`? The constructor takes the the RobotHW interface and would call the overridden `init()`.

hardware_interface/include/hardware_interface/robot_hw.h
```diff
@@ -51,8 +51,18 @@ namespace hardware_interface
  * hardware.
  *
  * The hardware interface map (\ref interfaces_) is a 1-to-1 map between
- * the names of interface types derived from \ref HardwareInterface  and
+ * the names of interface types derived from \ref HardwareInterface and
  * instances of those interface types.
+ * 
+ * The registration (\ref registerInterface) of interfaces can be done in the 
+ * constructor or init() of a custom robot hardware class.
```
fjp: As mentioned in [this comment](https://github.com/ros-controls/ros_control/pull/433#discussion_r403259048) I don't know where the interface registration should take place.

hardware_interface/include/hardware_interface/robot_hw.h
```diff
@@ -51,8 +51,18 @@ namespace hardware_interface
...
+ * 
+ * The robot state is described by data members of a class derived from this 
+ * base interface. The names of these data members should give semantic meaning 
+ * (pos, vel, eff) to the registered resources (joints, sensors, actuators). 
+ * Commands from controllers are represented by a member array (e.g. cmd)
+ * and populated by the controller's update method, 
+ * (controller_interface::ControllerBase::update()).
```
fjp: Not sure if this gives too much details on how to implement a derived class. If this is the wrong place, I think there should be a link to an example showing how it should be done. For example to the [wiki page](https://github.com/ros-controls/ros_control/wiki/hardware_interface#setting-up-a-new-robot). But as I've mentioned, I think this example should be updated.

hardware_interface/include/hardware_interface/robot_hw.h
```diff
@@ -70,6 +80,13 @@ class RobotHW : public InterfaceManager
 
   /** \brief The init function is called to initialize the RobotHW from a
    * non-realtime thread.
+   * 
+   * The registration of data members to interfaces is done via 
+   * \ref JointStateInterface, for read-only joints, or \ref JointCommandInterface, 
+   * for joints that accept commands and provide feedback (read and write), 
+   * including its derived classes (e.g. \ref VelocityJointInterface).
+   * These interfaces make use of \ref JointHandle or \ref JointStateHandle through 
+   * the \ref HardwareResourceManager.
```
fjp: Again, I am not sure if the registration of interfaces belongs to init or the constructor.

hardware_interface/include/hardware_interface/robot_hw.h
```diff
@@ -148,33 +168,66 @@ class RobotHW : public InterfaceManager
...
+   * and is used to populate the robot state from the robot's hardware resources
+   * (joints, sensors, actuators). This method should be called before 
+   * controller_manager::ControllerManager::update() and \ref write.
+   * 
+   * \note The name \ref read refers to reading state from the hardware.
+   * This complements \ref write, which refers to writing commands to the hardware.
+   *
+   * Querying WallTime inside \ref read is not realtime safe. The input parameters
+   * \ref time and \ref period make it possible to inject time from any 
+   * realtime backend framework.
```
fjp: I've added a reason why the `read` and `write` methods take `time` and `duration` parameters. I hope this is their intended use. Please let me know if not.

hardware_interface/include/hardware_interface/robot_hw.h
```diff
@@ -148,33 +168,66 @@ class RobotHW : public InterfaceManager
...
+   * 
+   * The write method is part of the control loop cycle (\ref read, update, \ref write) 
+   * and used to send out commands, via the hardware interfaces 
+   * (eg. JointCommandInterface and its subclasses), to the robot's hardware 
+   * resources (joints, actuators). Inside a node that handles the control loop, 
+   * this method should be called after \ref read and controller_manager::ControllerManager::update.
+   * 
+   * \note The name \ref write refers to writing (sending out) data commands,
+   * obtained from controllers, to the robot hardware resources.
+   * Use \ref read to write the state of the robot's hardware resources to data members.
```
fjp: Now this is also shorter, similar to read. Please let me know if it requires changes.

fjp:
> IMO the `\brief` on `prepareSwitch` and `doSwitch` doesn't add anything, and is too long to be a brief. I think it's fine without these changes, but if you want to give it a brief, then a shorter, less detailed summary might be better than just pulling the first sentence. Something like `Check (in non-realtime) if the given controllers can be started and stopped.`, and then leave the rest of the body as it was before to add the necessary details.
> 
> I added some suggestions below, everything I wrote for `read()` can be translated to `write()` as well. The suggestions probably violate the line length limits, please ignore my bad formatting :)
> 
> Overall I think it's great but the language can be simplified a lot. In particular, I'd like to stay away from getting too close to implementation details, since we shouldn't try to force a particular style of implementation. I'm also not a fan of "raw data", so in my suggestions I used "robot state" and "command" instead.

Thanks again for the review and your suggestions @matthew-reynolds. I've switched back the `\brief` onf both `prepareSwitch` and `doSwitch` and changed a lot on the other docstrings as well (not only on `read` and `write`).

hardware_interface/include/hardware_interface/robot_hw.h
matthew-reynolds: > Could you please clarify where the registration of interfaces should take place or is it just personal preference?

Off the top of my head, I can't think of a reason it would make a difference whether it's in the constructor or `init`, I've just always tended to put it in `init` :) Unless someone else chimes in with a good reason, I don't think it matters. I also don't think there's need to recommend one way vs the other, can we word the documentation to be inclusive of either option?

> Do you know why the init() method of ros_control_boilerplate isn't using the same parameters as the base class?

That looks like a mistake to me, but let's ping @davetcoleman to know for sure. Looks like a good example of a place where `override` would help.

> Wouldn't it be possible to move the call of `init(ros::NodeHandle& , ros::NodeHandle &)` inside the controller_manager constructor to avoid having the user calling init()?

Perhaps, but that might remove a bit of flexibility in case people are doing funky custom stuff. At this point in the lifecycle of `ros_control`, changing the API without a strong reason doesn't seem like a good idea to me.

hardware_interface/include/hardware_interface/robot_hw.h
matthew-reynolds: I think this flexibility is good, as I mentioned I don't think it matters where it takes place.

hardware_interface/include/hardware_interface/robot_hw.h
matthew-reynolds: IMO this is still too much detail about implementation, I don't really like recommending data types and variable names and that sort of thing, but mentioning that this class stores the state of the robot's hardware resources and outgoing commands is fine. Maybe it should live up in the first paragraph, to elaborate on "maintains a map of the hardware interfaces" (Line 49)?

hardware_interface/include/hardware_interface/robot_hw.h
matthew-reynolds: I think this should be reworded to be more generic - We are not limited to classes like `JointStateInterface`, `JointCommandInterface`, `JointHandle`, etc. Though these are the ones `ros_control` itself provides, users are completely free to define their own custom interfaces and handles.

hardware_interface/include/hardware_interface/robot_hw.h
matthew-reynolds: Yep, that's their intended use. Good addition.

A bit of a nitpick, but "input parameters" is a bit redundant, "parameters" is fine. I would also reword it to "inject time from a realtime source", but that might just be my personal writing preferences creeping in ;)

hardware_interface/include/hardware_interface/robot_hw.h
```diff
@@ -148,33 +168,66 @@ class RobotHW : public InterfaceManager
...
+   * realtime backend framework.
    *
    * \param time The current time
    * \param period The time passed since the last call to \ref read
    */
   virtual void read(const ros::Time& /*time*/, const ros::Duration& /*period*/) {}
 
-  /**
-   * Writes data to the robot HW
+  /** \brief Writes commands from controllers to the robot hardware.
```
matthew-reynolds: Let's make this present tense, and for consistency with `read()`, drop the "from controllers".

```suggestion
  /** \brief Write commands to the robot hardware.
```

hardware_interface/include/hardware_interface/robot_hw.h
```diff
@@ -148,33 +168,66 @@ class RobotHW : public InterfaceManager
...
+   * 
+   * The write method is part of the control loop cycle (read, update, write) 
+   * and used to send out commands, via the hardware interfaces 
+   * (eg. JointCommandInterface and its subclasses), to your robot's hardware 
+   * esources (joints, actuators). Inside a node that handles the control loop, 
+   * this method should be called after \ref read and controller_manager::ControllerManager::update.
```
matthew-reynolds: For consistency with `read()`, something like this:

```suggestion
   * The write method is part of the control loop cycle (\ref read, update, \ref write) 
   * and used to send out commands to the robot's hardware resources (joints, actuators).
   * This method should be called after \ref read and controller_manager::ControllerManager::update.
```

hardware_interface/include/hardware_interface/robot_hw.h
```diff
@@ -148,33 +168,66 @@ class RobotHW : public InterfaceManager
...
+   * 
+   * \note The name \ref write refers to writing (sending out) data commands,
+   * obtained from controllers, to the robot hardware resources.
+   * Use \ref read to write the state of the robot's hardware resources to data members.
```
matthew-reynolds: For consistency with `read()`:

```suggestion
   * \note The name \ref write refers to writing commands to the hardware.
   * This complements \ref read, which refers to reading state from the hardware.
```

hardware_interface/include/hardware_interface/robot_hw.h
fjp: I agree, this is probably not the best place for data type and variable name suggestions and I'll move this part up to the class description (Line 49). 

What about updating the [example in the wiki page](https://github.com/ros-controls/ros_control/wiki/hardware_interface#setting-up-a-new-robot)? I think this could be another option: we extend the example slightly (if required) and link to it in the class description (somewhere around line 49). Another way I could imagine is to put the wiki example into this class description. This is also done in the detailed description of the [transmission_interface](http://docs.ros.org/melodic/api/transmission_interface/html/c++/classtransmission__interface_1_1TransmissionInterfaceLoader.html#details).

hardware_interface/include/hardware_interface/robot_hw.h
fjp: Good point, I didn't think about custom ones. I'll try to find a more generic description.

hardware_interface/include/hardware_interface/robot_hw.h
fjp: Sounds nice :) Let's do it that way.

hardware_interface/include/hardware_interface/robot_hw.h
fjp: Looks good, I'll apply this in my next commit. Unfortunately I can't do it here on GitHub as your suggestion is marked outdated.

hardware_interface/include/hardware_interface/robot_hw.h
matthew-reynolds: @bmagyar Can you confirm what I said in my first point above about registering interfaces in the constructor vs init?

> Off the top of my head, I can't think of a reason it would make a difference whether it's in the constructor or init, I've just always tended to put it in init :) Unless someone else chimes in with a good reason, I don't think it matters. I also don't think there's need to recommend one way vs the other, can we word the documentation to be inclusive of either option?

hardware_interface/include/hardware_interface/robot_hw.h
matthew-reynolds: I like the idea of putting a minimal example (such as the wiki example you linked) in the class description, that's a lot less fragile than linking to the wiki since the wiki could always change and break the link.

hardware_interface/include/hardware_interface/robot_hw.h
```diff
@@ -148,33 +168,65 @@ class RobotHW : public InterfaceManager
...
   }
+  /**\}*/
 
-  /**
-   * Reads data from the robot HW
+  
+  /** \name Control Loop
+   * 
+   */
+   /**\{*/
```
matthew-reynolds: Should this be the same as the other groups? I'm not a Doxygen expert, just looking at consistency. Maybe this has something to do with the different grouping you see when you generate the doc? No idea :)

```suggestion
  /** \name Control Loop
   *\{*/
```

fjp: Thanks for the hint! Unfortunately it still looks different, not sure how to fix this.

fjp:
I have applied most of the suggestions so far but didn't put the example from the wiki page in the class description because I think it would overload the `hardware_interface::RobotHW` class. Instead, I've added a slightly modified version of it in a newly created `README.md` which is used as doxygen's mainpage and can also be used as a readme for the `hardware_interface` folder of the github repository itself. However, if the markdown formatting doesn't work it is also possible to create a new `mainpage.dox`, move the content of `README.md` there and just add a link to the `mainpage.dox` in the `README.md` and the delete the rest. One thing that is missing now is the `\htmlinclude manifest.html` which is usually found in a `mainpage.dox` of a ROS package. Note sure if this can be added inside the README.md. Please let me know what you think, any suggestions are welcome.

hardware_interface/include/hardware_interface/robot_hw.h
fjp: As mentioned below, I've put the wiki example (slightly modified) into a new `README.md` which acts as doxygen's mainpage and also as readme for the `hardware_interface` folder of the github repository.

hardware_interface/README.md
```diff
@@ -0,0 +1,299 @@
...
+    // Register the JointPositionInterface containing the read/write joints
+    // with this robot's hardware_interface::RobotHW.
+    registerInterface(&jnt_pos_interface);
+
+    return true;
+  }
+
+private:
+  hardware_interface::JointStateInterface jnt_state_interface;
+  hardware_interface::PositionJointInterface jnt_pos_interface;
```
fjp, line 104: Why exactly are both needed? Isn't [PositionJointInerface](http://docs.ros.org/melodic/api/hardware_interface/html/c++/classhardware__interface_1_1PositionJointInterface.html) sufficent because it inherits from [JointCommandInterface](http://docs.ros.org/melodic/api/hardware_interface/html/c++/classhardware__interface_1_1JointCommandInterface.html) which can be used for reading and writing? Or are both required because all [existing controllers](https://github.com/ros-controls/ros_controllers) require a `JointStateInterface` to read the state of the robot's resources and don't make use of read/write interfaces (only custom controllers could)?

hardware_interface/include/hardware_interface/robot_hw.h
```diff
@@ -80,6 +107,7 @@ class RobotHW : public InterfaceManager
    */
   virtual bool init(ros::NodeHandle& /*root_nh*/, ros::NodeHandle &/*robot_hw_nh*/) {return true;}
```
fjp, line 108: Is there a reason why all parameters are commented in the headers? I know only the type is required in the declaration of a method but I am asking because doxygen gave me warnings of the missing parameter names.

hardware_interface/include/hardware_interface/robot_hw.h
```diff
@@ -148,33 +178,59 @@ class RobotHW : public InterfaceManager
...
   virtual SwitchState switchResult(const ControllerInfo& /*controller*/) const
   {
     return DONE;
   }
+  /**\}*/
 
-  /**
-   * Reads data from the robot HW
+  /** \name Control Loop
+   *\{*/
```
fjp, line 195: I don't know how to make these groups look alike. `Hardware Interface Switching` seems to be more important to doxygen than `Resource Management` and `Control Loop`.

hardware_interface/rosdoc.yaml
```diff
@@ -1,5 +1,5 @@
 - builder: doxygen
   name: C++ API
   output_dir: c++
-  file_patterns: '*.c *.cpp *.h *.cc *.hh *.dox'
+  file_patterns: '*.c *.cpp *.h *.cc *.hh *.dox *.md'
```
fjp, line 4: This is required to use a mardown file (`README.md`) as doxygen's mainpage. See also [`rosdoc_lite`](http://wiki.ros.org/Doxygen#Embedding_Markdown_Pages) how to include markdown files. Only "problem" is that the `\htmlinclude manifest.html` is missing, which is usually present in default `mainpage.dox` of ROS packages.

hardware_interface/README.md
```diff
@@ -0,0 +1,299 @@
...
+
+hardware interfaces are used by ROS control in conjunction with one of the 
+available ROS controllers to send (\ref hardware_interface::RobotHW::write) 
+commands to the hardware and receive (\ref hardware_interface::RobotHW::read) 
+states from the robot's resources (joints, sensors, actuators).
+
+A list of available hardware interfaces (provided via the HardwareResourceManager) 
+as of this writing:
+
+- [JointCommandInterface](\ref hardware_interface::JointCommandInterface):
```
fjp, line 11: Not sure if these links will work in the README.md when viewed on the GitHub repo. It is working though in the generated doxygen documentation.

hardware_interface/include/hardware_interface/robot_hw.h
```diff
@@ -51,8 +51,26 @@ namespace hardware_interface
...
  * instances of those interface types.
+ * 
+ * A class derived from this base interface represents a robot and stores the 
+ * state of the robot's hardware resources (joints, sensors, actuators) and 
+ * outgoing commands in its data member arrays. The names of these members 
+ * should provide semantic meaning (e.g. pos, vel, eff, cmd). 
+ * Commands from controllers are populated by the controller's update method, 
+ * (controller_interface::ControllerBase::update()).
+ * For each resource a \ref JointStateHandle (for read only joints), 
+ * \ref JointHandle (for read and write joints) or custom handle can be used
```
fjp, line 64: This question is similar to the [previous one](https://github.com/ros-controls/ros_control/pull/433/files#r405166152). Examples I've seen so far use both (`JointStateHandle` and `JointHandle`) but wouldn't `JointHandle` be enough as it can be used for reading joint state and writing commands?

hardware_interface/README.md
```diff
@@ -0,0 +1,299 @@
...
+    while (ros::ok())
+    {
+        // Basic bookkeeping to get the system time in order to compute the control period.
+        const ros::Time     time   = ros:Time::now();
+        const ros::Duration period = time - prev_time;
+        
+        // Execution of the actual control loop.
+        robot.read();
+        cm.update(time, period);
+        root.write();
```
fjp, line 164: Where exactly in the control loop should we use the `transmission_interface` if a robot's transmissions need to be defined in software? Is this done inside `read()` and `write()` methods? Or is this already handled in the URDF definition (at least for the available transmissions such as [`SimpleTransmission`](http://docs.ros.org/melodic/api/transmission_interface/html/c++/classtransmission__interface_1_1SimpleTransmission.html)?

hardware_interface/README.md
```diff
@@ -0,0 +1,299 @@
...
+#include <hardware_interface/joint_state_interface.h>
+#include <hardware_interface/robot_hw.h>
+
+class MyRobot : public hardware_interface::RobotHW
+{
+public:
+  MyRobot() 
+  { 
+    // Initialization of the robot's resources (joints, sensors, actuators) and
+    // interfacs can be done here or inside init().
```
fjp, line 69: Maybe we could add here that it is possible to get the joint names from the URDF via the parameter server like the [ros_control_boilerplate/generic_hw_interface](https://github.com/PickNikRobotics/ros_control_boilerplate/blob/5e40c7e875f7ed58ed061f37d4898865e382170e/src/generic_hw_interface.cpp#L52) seems to do it (using [rosparam_shortcuts](https://github.com/PickNikRobotics/rosparam_shortcuts)).

hardware_interface/include/hardware_interface/robot_hw.h
matthew-reynolds: That's... A good question. I suspect that's a mistake. Commenting out unused parameters at the function definition is sometimes done to suppress compiler warnings, but I don't think there's any reason to comment them out in the declaration like this. Would you mind un-commenting them here in the header, and double-checking that this satisfies both Doxygen and the compiler? (ie there are no compiler warnings when you compile with `-Wunused-parameter`/`-Wunused`/`-Wextra`)

hardware_interface/include/hardware_interface/robot_hw.h
matthew-reynolds: Oh, I think I figured it out. The `Hardware Interface Switching` group contains the `SwitchState` enum, which isn't a member func, so the group gets displayed differently. Take a look at [this page of the Doxygen manual](http://www.doxygen.nl/manual/grouping.html#memgroup). near the bottom.

Not sure whether it's better to leave the enum as is, or move it out of the group so that it's more logical in the generated HTML... Let's get other opinions.

~In either case, let's update the group definitions anyways for consistency. From the manual, it looks like the name is supposed to be in a different comment than the `\{`, but we've been getting on fine with them in the same comment, so let's stick to that.~
Edit: Whoops, should have refreshed.

hardware_interface/include/hardware_interface/robot_hw.h
matthew-reynolds: See https://github.com/ros-controls/ros_control/pull/433#discussion_r405916399

matthew-reynolds:
Since PR this is growing larger, I just want to ping @ros-controls/ros_control for more opinions.

@fjp I will need a few days to read your newest changes :sweat_smile:

hardware_interface/include/hardware_interface/robot_hw.h
matthew-reynolds: See #399. Looks like these were probably extra accidental changes.

fjp: Thanks for the link. I've uncommented the parameters and compiled the `hardware_interface` tests (where `robot_hw.h` gets included) using: `catkin build -v --pre-clean --cmake-args -DCMAKE_C_FLAGS="-Wall -Wextra -Wunused-parameter -Wunused" --catkin-make-args tests` without any warnings. Should I commit these changes (uncommenting parameters) to this PR or in a separate one?

matthew-reynolds: Let's split it out, this is already going to be a fairly lengthy PR. Thanks :)

hardware_interface/README.md
matthew-reynolds: Looks like they do not work :(

https://github.com/fjp/ros_control/tree/melodic-devel/hardware_interface

matthew-reynolds:
@fjp Sorry to go back on you here, but I think this PR is losing focus a bit. Do you mind leaving the markdown group fixes, `\brief`s, and the `read()` and `write()` descriptions, but splitting the rest of this out into a new PR? I think those other changes need more discussion, particularly the inclusion of the big README.

The few changes I listed above are basically ready to be merged and address specific issues rather than general improvement of the documentation.

fjp:
@matthew-reynolds no worries, splitting this up is a good idea. I'll leave the changes you mention and move the large changes of the documentation (README) to a new PR.

fjp:
@matthew-reynolds I've reverted the large changes (remove README, revert rosdoc.yaml, remove details docstrings of RobotHW class and init method) of this PR in a new branch [extend-read-write-docstrings](https://github.com/fjp/ros_control/tree/extend-read-write-docstrings) in my fork. In this branch I've also squashed most of the commits as the commit messages were similar. Should I use this new branch for a new PR? 

Or do you prefer using this PR with one additional commit to `melodic-devel` branch that removes the README.md, reverts rosdoc.yaml and removes the added details of the RobotHW class description and init method?

matthew-reynolds:
I don't think it's a big deal either way. On one hand, if we stay here, we preserve the discussion. But on the other hand, it's getting hard to navigate this PR since there's so much discussion and back-and-forth. Plus we can (And should) link to this PR in the new one if we go that route.

Not sure if the higher-ups have an opinion one way or the other, but I'd say probably doesn't matter, up to you.

fjp:
Thank for the feedback @matthew-reynolds. As this PR has quite a lot of comments and is rather confusing I'd like to make a new PR on a new branch ([extend-read-write-docstrings](https://github.com/fjp/ros_control/tree/extend-read-write-docstrings)). Also I think it is cleaner if I start to work on branches and not on `melodic-devel`, even if it's a fork. 

Therefore, I created a new PR #444 which mainly takes care of the `read` and `write` method docstrings and the doxygen group names. For the rest of the changes in this PR (extend documentation of `hardware_interface`) I'll come up with a new PR once #444 get's sorted out. I hope this is ok?

bmagyar:
I have merged #444 and also picked it for Noetic. I'd recommend opening a new PR with 100-200 lines of unchallenged changes from this PR so we can guarantee a somewhat smooth review process.

fjp:
Great, thanks @bmagyar. I'll do my best to keep the changes in the upcoming PR small.

fjp:
This PR should become obsolete once #457 is merged.

hardware_interface/include/hardware_interface/robot_hw.h
davetcoleman: I do not recall why it is that way, probably a mistake from me or from the sources that inspired the boilerplate. Please make a PR to fix it :-)

bmagyar:
Closing on favour of #457 , thanks guys for the thorough discussion here

PR #436, opened 2020-04-09
Add xmlns for URDF, remove xmlns:xacro

matthew-reynolds:
Drops the unused `xmlns:xacro`, since we aren't using any `xacro`, and replaces it with the URDF ns.

Although not widely used (As far as I can see), technically URDF should specify an `xmlns` of `http://www.ros.org`, as specified by the `targetNamespace` of the [URDF schema](https://github.com/ros/urdfdom/blob/master/xsd/urdf.xsd). Might as well set a good example :)

---

Interesting note, this removes every instance of the string "xacro" in the repo.

### DIFF
--- a/transmission_interface/test/transmission_parser_test.cpp
+++ b/transmission_interface/test/transmission_parser_test.cpp
@@ -49,7 +49,7 @@ TEST(TransmissionParserTest, InvalidFile)
 TEST(TransmissionParserTest, EmptyFile)
 {
   TransmissionParser parser;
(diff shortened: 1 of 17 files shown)


bmagyar (approved):
Looks good to me although I wonder where the CI is :)

bmagyar:
Github-Travis integration is flaky again, build is fine: https://travis-ci.org/github/ros-controls/ros_control/builds/672780015

PR #438, opened 2020-04-09
[Noetic] Use CamelCase for ResourceManagerType typedef

matthew-reynolds:
All other classes and typedefs follow CamelCase, so just updating this to match.

Should target `noetic-devel` once we branch, since it breaks API.


matthew-reynolds:
Actually, I just realized this is in `hardware_interface/internal`. It may be safe for Melodic, since it only breaks "internal" API.

bmagyar (approved):

PR #440, opened 2020-04-14
Migrate to new industrial_ci

matthew-reynolds:
This is required for `industrial_ci`'s experimental Noetic support, and I'd also like to get it in before spending time on the codecov integration in `ros_controllers`. Plus, might as well stay up to date :)

Will duplicate over to the other repos (`ros_controllers`, `urdf_geometry_parser`, `realtime_tools`, etc) once approved.

See https://github.com/ros-industrial/industrial_ci/blob/master/doc/migration_guide.md

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 # Travis Continuous Integration Configuration File For ROS Control Projects
 # Author: Dave Coleman
-sudo: required
-dist: trusty
 language: generic
+services:
+  - docker
 
 notifications:
   email:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -23,6 +23,6 @@ env:
     - ROS_DISTRO=melodic ROS_REPO=ros-testing
 
 install:
-  - git clone https://github.com/ros-industrial/industrial_ci.git .ci_config
+  - git clone --quiet --depth 1 https://github.com/ros-industrial/industrial_ci.git .industrial_ci -b master
```
matthew-reynolds, line 24: Changed clone directory to `.industrial_ci` just to match the industrial_ci documentation, certainly doesn't make a big deal but I figured if people come looking to established repos such as this one for examples, it removes a friction point if we match the documentation as closely as is reasonable.

bmagyar: What attention to detail! Thanks!

bmagyar:
Thanks! Indeed I fired a quick PR yesterday to see if Noetic works but found that it wasn't and had to leave it at that

bmagyar (approved):

PR #441, opened 2020-04-15
Bump CMake version to avoid CMP0048 (Backport #427)

bmagyar:
Signed-off-by: Shane Loretz <sloretz@osrfoundation.org>


matthew-reynolds (approved):
:+1:

PR #450, opened 2020-04-23
Fix rqt displaying and handling of 'initialized' controllers

### LINKED ISSUE #445: [rqt_controller_manager] Cannot handle controller after loading it
rqt_controller_manager doesn't seem to handle controller loading correctly. After Loading a controller, the **state** column correctly switches to *initialized*, but the marker doesn't switch from grey to red and it is not possible to right-click on the controller to perform further switches.

rqt_controller_bug

When the controller gets switched directly to a different state using a service the gui updates correctly, but unloading and loading the controller again brings back the problem. *Load and Start* also works correctly.

I created a small reproduction repo at [rqt_controller_bug](https://github.com/RobertWilbrandt/rqt_controller_bug), but i also encountered this with different robot arms (franka panda, ur5) and some custom nodes. I tried to build all relevant projects directly from git in *melodic_devel*, you can find a wstool file in the linked repo. I think the problem appeared at the end of December 2019, but that is quite uncertain.

I am currently running on ROS melodic on Ubuntu 4.15.0-96-generic, but as this problem seems to appear for colleagues on different systems too that should not really matter too much.

I'll look into solving this and creating a PR by myself, but don't know when i will be able to do so.

RobertWilbrandt:
Fixes #445. This restores the rqt gui behavior before #391, which changed possible ControllerState states. I don't really understand enough about the other states (unknown, waiting, aborted) to know if any action for them is useful.

### DIFF
--- a/rqt_controller_manager/src/rqt_controller_manager/controller_manager.py
+++ b/rqt_controller_manager/src/rqt_controller_manager/controller_manager.py
@@ -103,7 +103,7 @@ class ControllerManager(Plugin):
         self._icons = {'running': QIcon(path + '/resource/led_green.png'),
                        'stopped': QIcon(path + '/resource/led_red.png'),
                        'uninitialized': QIcon(path + '/resource/led_off.png'),
-                       'initialized': QIcon(path + '/resource/led_off.png')}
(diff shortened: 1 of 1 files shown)


bmagyar (approved):
Looks good to me!

PR #456, opened 2020-05-14
Make handling of registered InterfaceManagers in InterfaceManager transparent

### LINKED ISSUE #452: Inconsistent handling of registered InterfaceManagers in InterfaceManager
Currently, the [hardware_interface::InterfaceManager](https://github.com/ros-controls/ros_control/blob/melodic-devel/hardware_interface/include/hardware_interface/internal/interface_manager.h) class allows to either register either interfaces directly or sub-```InterfaceManager```s (afaik only used for ```combined_robot_hw``` in this repo). But while the ```get<InterfaceType>()``` uses the registered ```InterfaceManager```s in order to find a valid interface/Resourcemanager, the ```getNames()``` and ```getInterfaceResources()``` ignore them completely. Because of this, when you use a ```RobotHW``` which registered another ```InterfaceManager```, you cannot use it in a ```combined_robot_hw``` (the filter-routines don't know about the 'transitively-registered' interfaces).

While it is probably not too tragic that you cannot use a ```combined_robot_hw``` inside another ```combined_robot_hw```, i had some siuations where using ```registerInterfaceManager()``` would have been quite handy (mostly wrapping existing ros-control drivers to be loadable in a ```combined_robot_hw```). The current state also seems pretty inconsistent to me, which is why i want to create a PR to handle all transitively registered interfaces consistently.

Because i don't know too much about other ros-control parts, i created this issue first before providing a PR myself. As far as i can tell this shouldn't break things inside ros_control, but please tell if this doesn't work for some reason.

RobertWilbrandt:
This fixes #452. Most importantly, it handles registered ```InterfaceManagers``` transparently inside the ```getNames()``` and ```getInterfaceResources()``` functions (adding interfaces and resources from them and removing duplicates when needed). In addition, a test for this functionality is added and the documentation is updated (most of it didn't get updated since when ```registerInterfaceManager()```was introduced). Finally, a small duplicate error message was removed.

### DIFF
--- a/hardware_interface/include/hardware_interface/internal/interface_manager.h
+++ b/hardware_interface/include/hardware_interface/internal/interface_manager.h
@@ -219,6 +219,19 @@ public:
     {
       out.push_back(interface.first);
     }
+
+    for (const auto& interface_manager : interface_managers_)
+    {
+      // Make sure interfaces appear only once, as they may have been combined
+      for (const auto& interface_name : interface_manager->getNames())
+      {
+        if (std::find(out.begin(), out.end(), interface_name) == out.end())
+        {
+          out.push_back(interface_name);
+        }
+      }
+    }
+
     return out;
   }
 
@@ -237,6 +250,13 @@ public:
     {
       out = it->second;
     }
+
+    for (const auto& interface_manager : interface_managers_)
+    {
+      std::vector<std::string> resources = interface_manager->getInterfaceResources(iface_type);
+      out.insert(out.end(), resources.begin(), resources.end());
+    }
+
     return out;
   }
 
--- a/hardware_interface/include/hardware_interface/internal/interface_manager.h
+++ b/hardware_interface/include/hardware_interface/internal/interface_manager.h
@@ -110,6 +110,15 @@ struct CheckIsResourceManager {
 
 } // namespace internal
 
+/**
+ * \brief Manager for hardware interface registrations.
+ *
+ * This class enables the registration of interfaces based on their class type,
+ * handling all the required demangling and storage. The registration ensures
+ * the presence of at most one interface instance per type. Accessors for
+ * interface listing are provided. Additionally, combinations of interfaces as
+ * required in \c CombinedRobotHW are handled transparently.
+ */
 class InterfaceManager
 {
 public:
@@ -118,6 +127,9 @@ public:
    *
    * This associates the name of the type of interface to be registered with
    * the given pointer.
+   *
+   * \note The registration of an interface will replace previously registered
+   * instances of its type
    *
    * \tparam T The interface type
    * \param iface A pointer to the interface to store
@@ -134,6 +146,14 @@ public:
     internal::CheckIsResourceManager<T>::callGetResources(resources_[iface_name], iface);
   }
 
+  /**
+   * \brief Register another interface manager.
+   *
+   * This manager will be integrated transparently into all further access
+   * methods.
+   *
+   * \param iface_man A pointer to the interface manager to store
+   */
   void registerInterfaceManager(InterfaceManager* iface_man)
   {
     interface_managers_.push_back(iface_man);
@@ -142,12 +162,23 @@ public:
   /**
    * \brief Get an interface.
    *
-   * Since this class only stores one interface per type, this returns a
-   * pointer to the requested interface type. If the interface type is not
-   * registered, it will return \c NULL.
+   * If this class and its registered sub-managers only have one registered
+   * instance of the requested type, this will be returned. If multiple
+   * instances of the interface type were registered and the type is a
+   * \c ResourceManager, this call will try to combine them into a single
+   * handle. In all other cases, \c nullptr will be returned.
+   *
+   * \note As this instance will only store one instance of each manager type,
+   * the only way multiple interfaces are registered is the registration of
+   * other managers with interfaces of the same type.
+   *
+   * \note If there are multiple registered interfaces of the requested type
+   * (either directly or through registered interfaces) and they are not
+   * \c ResourceManager, this method will not be able to combine them and will
+   * return \c nullptr.
    *
    * \tparam T The interface type
-   * \return A pointer to the stored interface of type \c T or \c NULL
+   * \return A handle for the stored interfaces of type \c T or \c nullptr
    */
   template<class T>
   T* get()
@@ -208,7 +239,15 @@ public:
     return iface_combo;
   }
 
-  /** \return Vector of interface names registered to this instance. */
+  /**
+   * \brief Lists the demangled names of all registered interfaces.
+   *
+   * This includes the interfaces directly registered to this instance and
+   * the interfaces registered to registered managers. As multiple interfaces
+   * of the same type will get merged on access, duplicates are omitted.
+   *
+   * \return Vector of demangled interface names of registered interfaces.
+   **/
   std::vector<std::string> getNames() const
   {
     std::vector<std::string> out;
@@ -234,10 +273,12 @@ public:
   }
 
   /**
-   * \brief Get the resource names registered to an interface, specified by type
-   * (as this class only stores one interface per type)
+   * \brief Get the resource names registered to an interface, specified by type.
    *
-   * \param iface_type A string with the demangled type name of the interface
+   * This will return the list of all registered resources for
+   * \c ResourceManager interfaces and an empty list for all others.
+   *
+   * \param iface_type The demangled type name of an interface
    * \return A vector of resource names registered to this interface
    */
   std::vector<std::string> getInterfaceResources(std::string iface_type) const
--- a/hardware_interface/include/hardware_interface/internal/interface_manager.h
+++ b/hardware_interface/include/hardware_interface/internal/interface_manager.h
@@ -200,10 +200,8 @@ public:
         interfaces_combo_[type_name] = iface_combo;
         num_ifaces_registered_[type_name] = iface_list.size();
       } else {
-        // it is not a ResourceManager
-        ROS_ERROR("You cannot register multiple interfaces of the same type which are "
-                  "not of type ResourceManager. There is no established protocol "
-                  "for combining them.");
+        // Multiple interfaces of the same type which are not ResourceManager cannot be combined, so return
+        // nullptr
         iface_combo = nullptr;
       }
     }
--- a/hardware_interface/include/hardware_interface/internal/interface_manager.h
+++ b/hardware_interface/include/hardware_interface/internal/interface_manager.h
@@ -163,19 +163,18 @@ public:
    * \brief Get an interface.
    *
    * If this class and its registered sub-managers only have one registered
-   * instance of the requested type, this will be returned. If multiple
-   * instances of the interface type were registered and the type is a
+   * instance of the requested interface type, this will be returned. If
+   * multiple instances of the interface type were registered and the type is a
    * \c ResourceManager, this call will try to combine them into a single
    * handle. In all other cases, \c nullptr will be returned.
    *
-   * \note As this instance will only store one instance of each manager type,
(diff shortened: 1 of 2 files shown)


hardware_interface/include/hardware_interface/internal/interface_manager.h
```diff
@@ -219,14 +247,29 @@ class InterfaceManager
...
+
+    for (const auto& interface_manager : interface_managers_)
+    {
+      // Make sure interfaces appear only once, as they may have been combined
+      for (const auto& interface_name : interface_manager->getNames())
+      {
+        if (std::find(out.begin(), out.end(), interface_name) == out.end())
+        {
+          out.push_back(interface_name);
+        }
```
matthew-reynolds, line 267: If only we could safely modify the return type to be a `std::set` and then elegantly merge the sets from each sub-manager. Oh well, since that's not possible, this is good. No change requested, just thinking out loud :)

hardware_interface/test/interface_manager_test.cpp
```diff
@@ -84,6 +84,93 @@ TEST(InterfaceManagerTest, InterfaceRewriting)
   EXPECT_EQ(2, foo_iface_ptr->foo);
 }
 
+struct TestCombinationHandle {
+  TestCombinationHandle(const std::string& name) : name{name} {}
```
matthew-reynolds: Nit: brace on newline and prefer parens over braces for init

```suggestion
struct TestCombinationHandle
{
  TestCombinationHandle(const std::string& name) : name(name) {}
```

hardware_interface/test/interface_manager_test.cpp
```diff
@@ -84,6 +84,93 @@ TEST(InterfaceManagerTest, InterfaceRewriting)
   EXPECT_EQ(2, foo_iface_ptr->foo);
 }
 
+struct TestCombinationHandle {
```
matthew-reynolds: Should be class since it contains member funcs

hardware_interface/test/interface_manager_test.cpp
```diff
@@ -84,6 +84,93 @@ TEST(InterfaceManagerTest, InterfaceRewriting)
...
+    return name;
+  }
+
+  std::string name;
+};
+
+struct TestCombinationInterface : public ResourceManager<TestCombinationHandle>
+{
+
+};
```
matthew-reynolds: Let's make this a class and use the same formatting for the other Interfaces:

```suggestion
class TestCombinationInterface : public ResourceManager<TestCombinationHandle> {};
```

hardware_interface/test/interface_manager_test.cpp
matthew-reynolds: Nit: I think these two struct/classes should be above all the tests (ie. line 50). Group all the type definitions, and group all the tests.

hardware_interface/test/interface_manager_test.cpp
```diff
@@ -84,6 +84,93 @@ TEST(InterfaceManagerTest, InterfaceRewriting)
...
+
+  TestCombinationInterface combi_root_iface;
+  TestCombinationHandle combi_root_handle("foo_handle");
+  combi_root_iface.registerHandle(combi_root_handle);
+  root_mgr.registerInterface(&combi_root_iface);
+
+  TestCombinationInterface combi_leaf_iface;
+  TestCombinationHandle combi_leaf_handle("bar_handle");
+  combi_leaf_iface.registerHandle(combi_leaf_handle);
+  leaf_mgr.registerInterface(&combi_leaf_iface);
```
matthew-reynolds: Can you rename these handles from `"foo_handle"` and `"bar_handle"`? I think it causes confusion with the `FooInterface` above. Maybe just match the var name, so `"combi_root_handle"` and `"combi_leaf_handle"`? Removes ambiguity and also adds "root" and "leaf" to the names which is helpful

hardware_interface/test/interface_manager_test.cpp
```diff
@@ -84,6 +84,93 @@ TEST(InterfaceManagerTest, InterfaceRewriting)
...
+
+  // BazInterface should still work, as there is only one interface of this type
+  BazInterface* baz_iface_ptr = root_mgr.get<BazInterface>();
+  EXPECT_NE(nullptr, baz_iface_ptr);
+  EXPECT_EQ(4.2, baz_iface_ptr->baz);
+
+  // Check presence of handles in combined interface
+  TestCombinationInterface* combi_iface_ptr = root_mgr.get<TestCombinationInterface>();
+  std::vector<std::string> combi_handle_names = combi_iface_ptr->getNames();
+  EXPECT_NE(nullptr, combi_iface_ptr);
```
matthew-reynolds, line 146: Check `combi_iface_ptr` before dereferencing it, otherwise the test could crash in a way that doesn't give us a useful failure message

hardware_interface/include/hardware_interface/internal/interface_manager.h
```diff
@@ -199,18 +225,20 @@ class InterfaceManager
         // save the combined interface for if this is called again
         interfaces_combo_[type_name] = iface_combo;
         num_ifaces_registered_[type_name] = iface_list.size();
-      } else {
-        // it is not a ResourceManager
-        ROS_ERROR("You cannot register multiple interfaces of the same type which are "
-                  "not of type ResourceManager. There is no established protocol "
-                  "for combining them.");
-        iface_combo = nullptr;
```
matthew-reynolds: Rather than deleting this altogether, can you leave the explicit (redundant) `iface_combo = nullptr` and change the `ROS_ERROR` to a comment saying something along the lines of "Multiple interfaces of the same type which are not ResourceManager cannot be combined, so return nullptr".

I think that that makes it clearer how this condition is handled.

hardware_interface/include/hardware_interface/internal/interface_manager.h
```diff
@@ -142,12 +162,18 @@ class InterfaceManager
...
-   * registered, it will return \c NULL.
+   * If this class and its registered sub-managers only have one registered
+   * instance of the requested type, this will be returned. If multiple
+   * instances of the interface type were registered and the type is a
+   * \c ResourceManager, this call will try to combine them into a single
+   * handle. In all other cases, \c nullptr will be returned.
+   *
+   * \note As this instance will only store one instance of each manager type,
+   * the only way multiple interfaces are registered is the registration of
+   * other managers with interfaces of the same type.
```
matthew-reynolds: Please add a note to this explaining that, in this case, this function will return null because there is no method of combining these non-`ResourceManager` interfaces.

I just want to make sure it's very clear in the documentation that this function has a shortcoming under that condition, since it seems to me that that condition (multiple interfaces of the same type which isn't `ResourceManager`) is quite possible. To keep API compatible, returning nullptr is the best we can do, but it does seem like a shortcoming over a more robust solution like returning multiple interfaces or returning the directly-managed interface over sub-manager interfaces or something like that.

matthew-reynolds (changes requested):
I don't work with `combined_robot_hw` so I'm not totally familiar with it or with the use-cases for this PR, but from your description in #452, this seems like a very safe and useful change.

Couple small requested changes, but overall looks great. Big +1 for adding tests. @bmagyar I would definitely like to hear your thoughts since you are much more familiar with the combined robot hw than I.

hardware_interface/include/hardware_interface/internal/interface_manager.h
RobertWilbrandt: Agreed, that would make it all neater

hardware_interface/test/interface_manager_test.cpp
RobertWilbrandt: Applied in rebased commits

hardware_interface/test/interface_manager_test.cpp
RobertWilbrandt: Applied in rebased commits

hardware_interface/test/interface_manager_test.cpp
RobertWilbrandt: Applied in rebased commits

hardware_interface/test/interface_manager_test.cpp
RobertWilbrandt: Applied in rebased commits

hardware_interface/test/interface_manager_test.cpp
RobertWilbrandt: Agreed, that makes much more sense. Applied in rebased commits.

hardware_interface/test/interface_manager_test.cpp
RobertWilbrandt: Nice catch, applied in rebased commits

hardware_interface/include/hardware_interface/internal/interface_manager.h
RobertWilbrandt: Sure, that seems reasonable (i don't really have to make navigating these files any harder...). Applied in rebased commits.

hardware_interface/include/hardware_interface/internal/interface_manager.h
RobertWilbrandt: Added in rebased commits

RobertWilbrandt:
Thanks for looking through this. I tried applying all requested changes, and since there was some time since the initial pr i went ahead and rebased to current ```melodic_devel```.

hardware_interface/include/hardware_interface/internal/interface_manager.h
```diff
@@ -142,12 +162,23 @@ class InterfaceManager
...
```
matthew-reynolds: Small language clarification. (Not sure if this suggestion respects line lengths, some reformatting may be needed)

```suggestion
   * If this class and its registered sub-managers only have one registered
   * instance of the requested interface type, this will be returned. If multiple
   * instances of the interface type were registered and the type is a
   * \c ResourceManager, this call will try to combine them into a single
   * handle. In all other cases, \c nullptr will be returned.
   *
   * \note As this class can only store one instance of each interface type,
   * the only way multiple instances of the same interface type can be
    * registered is through the registration of sub-managers.
```

hardware_interface/include/hardware_interface/internal/interface_manager.h
```diff
@@ -142,12 +162,23 @@ class InterfaceManager
...
+   *
+   * \note If there are multiple registered interfaces of the requested type
+   * (either directly or through registered interfaces) and they are not
```
matthew-reynolds: Not sure this "(either directly or through registered interfaces)" is needed. If anything, it should say "registered sub-managers", but I think the rest of this block comment describes the need for sub-managers already.

matthew-reynolds (approved):
LGTM, couple documentation notes but overall looks great. Thanks for the rebase

hardware_interface/include/hardware_interface/internal/interface_manager.h
RobertWilbrandt: Sure, applied in 04bcb42a80da3b7f6e8dc498532723ab8ba00fee

hardware_interface/include/hardware_interface/internal/interface_manager.h
RobertWilbrandt: Sure, applied in 04bcb42a80da3b7f6e8dc498532723ab8ba00fee

RobertWilbrandt:
I applied the documentation changes. Regarding the formatting: Are there any guidelines for this? I tried keeping line-lengths etc consistent for the touched files, but other files are formatted completely different. Is there any preferred clang-format config (e.g. moveit)?

bmagyar:
It actually looks all good to me I was just hoping for a little more feedback from others. Let's go ahead with this though, I'll cherry-pick it to Noetic as well and push some releases

PR #457, opened 2020-05-16
[hardware_interface] Update documentation

fjp:
This PR is split from #433. It adds a `mainpage.dox` including the text and examples from the [hardware_interface GitHub wiki page](https://github.com/ros-controls/ros_control/wiki/hardware_interface#setting-up-a-new-robot) and the [presentation](https://vimeo.com/107507546) of @adolfo-rt. I've also updated the docstrings of class `RobotHW` and its `init()` method.

Here are the things that weren't fully discussed in #433:
- Where should registering of interfaces take place? Constructor, init or doesn't matter?
- Inside the example we could also mention that it is possible to get the joint names from the URDF via the parameter server like the [ros_control_boilerplate/generic_hw_interface](https://github.com/PickNikRobotics/ros_control_boilerplate/blob/5e40c7e875f7ed58ed061f37d4898865e382170e/src/generic_hw_interface.cpp#L52) seems to do it (using [rosparam_shortcuts](https://github.com/PickNikRobotics/rosparam_shortcuts)).
- In the example from the wiki: Why are both JointStateInterface and PositionJointInterface members needed? Isn't [PositionJointInerface](http://docs.ros.org/melodic/api/hardware_interface/html/c++/classhardware__interface_1_1PositionJointInterface.html) sufficent because it inherits from [JointCommandInterface](http://docs.ros.org/melodic/api/hardware_interface/html/c++/classhardware__interface_1_1JointCommandInterface.html) which can be used for reading and writing? Or are both required because all [existing controllers](https://github.com/ros-controls/ros_controllers) require a `JointStateInterface` to read the state of the robot's resources and don't make use of read/write interfaces (only custom controllers could)?
- Where exactly in the control loop should we use the `transmission_interface` if a robot's transmissions need to be defined in software? Is this done inside `read()` and `write()` methods? Or is this already handled in the URDF definition (at least for the available transmissions such as [`SimpleTransmission`](http://docs.ros.org/melodic/api/transmission_interface/html/c++/classtransmission__interface_1_1SimpleTransmission.html)?
- The method parameters are commented in the headers which is a result from #399. This gives warnings when building the documentation with `rosdoc_lite`. I'll address this in a new PR after this one is merged.


hardware_interface/mainpage.dox
```diff
@@ -0,0 +1,303 @@
...
+#include <hardware_interface/joint_state_interface.h>
+#include <hardware_interface/robot_hw.h>
+
+class MyRobot : public hardware_interface::RobotHW
+{
+public:
+  MyRobot() 
+  { 
+    // Initialization of the robot's resources (joints, sensors, actuators) and
+    // interfaces can be done here or inside init().
```
fjp, line 78: Maybe we could add here that it is possible to get the joint names from the URDF via the parameter server like the [ros_control_boilerplate/generic_hw_interface](https://github.com/PickNikRobotics/ros_control_boilerplate/blob/5e40c7e875f7ed58ed061f37d4898865e382170e/src/generic_hw_interface.cpp#L52) seems to do it (using [rosparam_shortcuts](https://github.com/PickNikRobotics/rosparam_shortcuts)).

hardware_interface/mainpage.dox
```diff
@@ -0,0 +1,303 @@
...
+    // Register the JointPositionInterface containing the read/write joints
+    // with this robot's hardware_interface::RobotHW.
+    registerInterface(&jnt_pos_interface);
+
+    return true;
+  }
+
+private:
+  hardware_interface::JointStateInterface jnt_state_interface;
+  hardware_interface::PositionJointInterface jnt_pos_interface;
```
fjp: Why exactly are both needed? Isn't [PositionJointInerface](http://docs.ros.org/melodic/api/hardware_interface/html/c++/classhardware__interface_1_1PositionJointInterface.html) sufficent because it inherits from [JointCommandInterface](http://docs.ros.org/melodic/api/hardware_interface/html/c++/classhardware__interface_1_1JointCommandInterface.html) which can be used for reading and writing? Or are both required because all [existing controllers](https://github.com/ros-controls/ros_controllers) require a `JointStateInterface` to read the state of the robot's resources and don't make use of read/write interfaces (only custom controllers could)?

hardware_interface/mainpage.dox
```diff
@@ -0,0 +1,303 @@
...
+    while (ros::ok())
+    {
+        // Basic bookkeeping to get the system time in order to compute the control period.
+        const ros::Time     time   = ros:Time::now();
+        const ros::Duration period = time - prev_time;
+        
+        // Execution of the actual control loop.
+        robot.read();
+        cm.update(time, period);
+        root.write();
```
fjp: Where exactly in the control loop should we use the `transmission_interface` if a robot's transmissions need to be defined in software? Is this done inside `read()` and `write()` methods? Or is this already handled in the URDF definition (at least for the available transmissions such as [`SimpleTransmission`](http://docs.ros.org/melodic/api/transmission_interface/html/c++/classtransmission__interface_1_1SimpleTransmission.html)?

hardware_interface/include/hardware_interface/robot_hw.h
```diff
@@ -54,6 +54,24 @@ namespace hardware_interface
...
  * instances of those interface types.
  *
+ * A class derived from this base interface represents a robot and stores the
+ * state of the robot's hardware resources (joints, sensors, actuators) and
+ * outgoing commands in its data member arrays. The names of these members
+ * should provide semantic meaning (e.g. pos, vel, eff, cmd).
+ * Commands from controllers are populated by the controller's update method,
+ * (controller_interface::ControllerBase::update()).
+ * For each resource a \ref JointStateHandle (for read only joints),
+ * \ref JointHandle (for read and write joints) or custom handle can be used
```
fjp, line 64: This question is similar to [this one](https://github.com/ros-controls/ros_control/pull/457/files#r426170320). Examples I've seen so far use both (`JointStateHandle` and `JointHandle`) but wouldn't `JointHandle` be enough as it can be used for reading joint state and writing commands?

hardware_interface/mainpage.dox
bmagyar: `JointStateInterface` gives you read access to all joint values without conflicting with other controllers. If you were to use a PositionJointInterface but only ever intended it for reading positions from it, other controllers can't write to it either as you reserved it for writing with this interface.

hardware_interface/include/hardware_interface/robot_hw.h
bmagyar: `JointStateHandle` doesn't create a conflict between controllers as it is read-only.

hardware_interface/mainpage.dox
bmagyar:
```suggestion
    // Initialization of the robot's resources (joints, sensors, actuators) and
    // interfaces can be done here or inside init().
    // E.g. parse the URDF for joint names & interfaces, then initialize them
```

hardware_interface/mainpage.dox
bmagyar: You'd need to call the right `propagate()` methods after your read and before your write

http://docs.ros.org/melodic/api/transmission_interface/html/c++/index.html, there's an example at the bottom of this page. I wouldn't add that to this example but probably link to this?

bmagyar (changes requested):

bmagyar:
Hi @fjp , I've added some small notes but in general I think this is 95% ready to merge. That 5% is made up of my comments ;)

hardware_interface/mainpage.dox
fjp: Thanks for explaining this. Then I think I'll add a comment describing this and link to the example.

fjp:
Thanks @bmagyar for your review and the helpful comments :+1: I'll try to find some time to finish the last 5% as soon as possible.

fjp:
@bmagyar I've addressed your comments. Please let me know if something is missing or needs changes. Thanks.

bmagyar (approved):
Thanks very much for this, you've saved a lot of headache for a lot of future ROS developers :)

Issue #443, opened 2020-04-17
Controller Manager Freezing on multiple controller switches


krishnachaitanya7:
My PC Specifications:
OS: Pop!_OS 18.04 LTS x86_64 
Host: Razer Blade 3.04 
CPU: Intel i7-9750H (12) @ 4.500GHz
GPU: NVIDIA NVIDIA Corporation Device RTX 2070
RAM: 16GB
ROS Version: Melodic

I have a gazebo simulation of Franka Emika robot, and when I switch the controllers multiple times it freezes. When I turn on the DEBUG statements It prints debug statement at this line: https://github.com/ros-controls/ros_control/blob/melodic-devel/controller_manager/src/controller_manager.cpp#L610 and doesn't move forward from there. I guess it's getting stuck in the while loop in line 611. I don't know what might be causing this problem. I don't know what extra details to post, but I will post whatever you guys ask for. Thank you!

bmagyar:
Is this sim publicly available? If yes, could you please post some steps to reproduce the issue?

krishnachaitanya7:
Hi @bmagyar. Below are the steps to reproduce my problem using the simulation I use.
```bash
mkdir freezing_issue && cd freezing_issue
mkdir src && cd src
git clone --single-branch --branch AddFreezingScript https://github.com/HIRO-group/ros_robotic_skin.git
cd ros_robotic_skin
./install.sh
```
Kindly note that we will be using the `AddFreezingScript` branch rather than master branch here.

The freezing issue would be the catkin workspace name. Once you source /PATH/TO/freezing_issue/devel/setup.bash by adding it to your bashrc (This step is basically activating the catkin workspace devel), you can run the simulation using:
```bash
roslaunch ros_robotic_skin simulation.launch
```
Now a gazebo window should open up with the simulation of franka panda robot. Now to reproduce the actual freezing you need to run:
```bash
rosrun ros_robotic_skin infinite_switches.py
```
We have created some controllers like Position Controllers and Velocity Controllers, and we need to repeatedly switch in between them. The `infinite_switches.py` accurately simulates the problem I am going through. 
Kindly let me know if I face some issues reproducing the problem, and I will try to answer right away. Thank you so much for your help.

krishnachaitanya7:
Hi, @bmagyar kindly lemme know if I can be of any help in reproducing the issue. I will attend to it right away. Thanks!

Issue #265, opened 2017-02-25
Multiple controller_manager spawn calls freeze a controller manager


mrjogo:
Using https://github.com/davetcoleman/ros_control_boilerplate/tree/0.4.0/rrbot_control as an example, replace the `ros_control_controller_manager` node in https://github.com/davetcoleman/ros_control_boilerplate/blob/0.4.0/rrbot_control/launch/rrbot_hardware.launch#L21 with:

```xml
<node name="ros_control_controller_manager" pkg="controller_manager" type="controller_manager" respawn="false"
	  output="screen" args="spawn joint_state_controller" />
<node name="ros_control_controller_manager" pkg="controller_manager" type="controller_manager" respawn="false"
	  output="screen" args="spawn position_trajectory_controller" />
```

Build and run `roslaunch rrbot_control rrbot_hardware.launch` and the controller manager stops responding, either to service calls (ie, `/rrbot/controller_manager/list_controllers`) or Ctrl-C (if run outside roslaunch). The same thing happens if using spawner instead of controller_manager. I haven't tested it on a controller manager running outside of ros_control_boilerplate.

If it's intended that only one spawn call is allowed, I didn't see it in the documentation.

bmagyar:
Your snippet clearly launches 2 controller managers so that's where the problem is.
Starting up 2 controller_manager nodes which lead to a resource and name conflict. The reason it stops responding is probably because either something gets stuck or the first controller manager dies when the second one is started.

Take a look at this [example](https://github.com/pal-robotics/tiago_simulation/blob/indigo-devel/tiago_controller_configuration_gazebo/launch/default_controllers.launch#L7) where the `spawner` script is used to set up controllers on an already running `controller_manager`. If you intend to start things up separately or in a parametrised way, this is the way to go.

`spawner` is only a script that calls services `load_controller`, `switch_controller`. It has the same effect as you'd get by executing the same commands from the command line, see [here](https://github.com/ros-controls/ros_control/blob/kinetic-devel/controller_manager/scripts/spawner#L115).

Dave's boilerplate functionalities allow a quick setup by using the controller manager + spawner at the same time but if you need to launch things in a modular way I'd recommend looking at any of the PAL robots for inspiration.

mathias-luedtke:
> Your snippet clearly launches 2 controller managers so that's where the problem is.

I think the problem is that both use the same node name `ros_control_controller_manager`.
So the first might get killed in the middle of the service call.
Please try it with distinct names.

[...]

PR #458, opened 2020-05-25
Adding extra time counter to exit while loop so that controller manager doesn't freeze

krishnachaitanya7:
This PR addresses issue #443. When I switch the controllers sometimes the controller manager is stuck [here](https://github.com/krishnachaitanya9/ros_control/blob/melodic-devel/controller_manager/src/controller_manager.cpp#L609) at this unending while loop. The problem I am facing is similar to #265. The only way out of this loop is either when 
- switch_params_.do_switch is False 
- ROS core exits 
- when there is a timeout. 

In the first case, switch_params_.do_switch becomes False if switching is successful which is the ideal case

Exiting ROS Core means either simulation is shutdown or ROS core exited with non zero exit code

If the timeout is specified after the timeout the switching is stopped and the response code returned back is False. The timeout is calculated using the ROS time. When I ran my simulation using Gazebo the ROS time freezes hence the timeout is never possible. Due to this issue,  the code is stuck at the aforementioned while loop and never comes out. This gives rise to Denial of Service. The simulation seems to be stuck. No further switching of controllers happen. Even the controller manager does not list the current controllers which are stopped or running because the controller manager is in denial of service and it's lock is not released. The only way out of this is to restart the entire simulation.  

Hence I have added a time check in the while loop which uses system time contrary to ROS time and if something like this happens at least the code can safely exit rather than go into denial of service. The code edits suggested in this PR will give every opportunity for ROS time to function but it will also monitor the timeout from a system clock perspective and after timeout+1 seconds if the while loop didn't break, it will forcibly break out of the while loop. Finally when the while loop is exited by forcibly breaking out "False" is returned signaling the end-user that controller switching wasn't successful.

### DIFF
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -613,7 +613,7 @@ bool ControllerManager::switchController(const std::vector<std::string>& start_c
       return false;
     }
     std::chrono::duration<double> diff = std::chrono::system_clock::now() - start_time;
-    if (diff.count() < timeout || timeout == 0){
+    if (diff.count() < timeout){
         std::this_thread::sleep_for(std::chrono::microseconds(100));
     } else {
         ROS_DEBUG("Time Out while Switching the controllers. Exiting cleanly");
--- a/controller_manager/src/controller_manager.cpp
+++ b/controller_manager/src/controller_manager.cpp
@@ -604,17 +604,29 @@ bool ControllerManager::switchController(const std::vector<std::string>& start_c
 
   // wait until switch is finished
   ROS_DEBUG("Request atomic controller switch from realtime loop");
+  auto start_time = std::chrono::system_clock::now();
+  bool timed_out = false;
   while (ros::ok() && switch_params_.do_switch)
   {
     if (!ros::ok())
     {
       return false;
     }
-    std::this_thread::sleep_for(std::chrono::microseconds(100));
+    std::chrono::duration<double> diff = std::chrono::system_clock::now() - start_time;
+    if (diff.count() < timeout || timeout == 0){
+        std::this_thread::sleep_for(std::chrono::microseconds(100));
+    } else{
+        ROS_DEBUG("Time Out while Switching the controllers. Exiting cleanly");
+        timed_out = true;
+        break;
+    }
+
   }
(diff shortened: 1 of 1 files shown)


krishnachaitanya7:
Hi @bmagyar! Can you please look into the PR? You commented on the previous issue I posted which I fixed in this PR. Thank you. Kindly let me know!

peasant98:
@bmagyar can you let us know the status of this? This is a known bug in the controller manager.

peasant98:
@bmagyar -- still no updates on this. Have you been able to verify that this bug occurs on your end?

controller_manager/src/controller_manager.cpp
```diff
@@ -604,17 +604,29 @@ bool ControllerManager::switchController(const std::vector<std::string>& start_c
...
     if (!ros::ok())
     {
       return false;
     }
-    std::this_thread::sleep_for(std::chrono::microseconds(100));
+    std::chrono::duration<double> diff = std::chrono::system_clock::now() - start_time;
+    if (diff.count() < timeout+1.0 || timeout == 0){
+        std::this_thread::sleep_for(std::chrono::microseconds(100));
+    } else {
+        ROS_DEBUG("Time Out while Switching the controllers. Exiting cleanly");
```
bmagyar:
```suggestion
        ROS_DEBUG("Timed out while switching controllers. Exiting...");
```

controller_manager/src/controller_manager.cpp
```diff
@@ -604,17 +604,29 @@ bool ControllerManager::switchController(const std::vector<std::string>& start_c
...
+        timed_out = true;
+        break;
+    }
   }
   start_request_.clear();
   stop_request_.clear();
-
+  if(timed_out){
+      ROS_DEBUG("Exited while loop forcibly without ROS handling timeout");
```
bmagyar:
```suggestion
      ROS_DEBUG("Exited wait until switch is finished loop using non-ROS-time timeout");
```

bmagyar:
Thanks for the fix, I'm releasing this now to melodic and noetic!

krishnachaitanya7:
Thank you so much @bmagyar!

PR #470, opened 2020-07-30
Remove ParseStatus enum

### LINKED ISSUE #333: #define in system header netdb.h conflicts with transmission loader enum
For some context, i'm trying to use boost asio in conjunction with the ros control RobotHw and there are of course transmissions involved. In trying to use both `transmission_interface` and `boost::asio`(via the `#include <boost/asio.hpp>` header), I have discovered a conflict between a `#define` in `netdb.h` which is used by boost and an enum defined in `transmission_loader.h`. Long story short if you have Boost installed there'll be a dependence on `netdb.h` and in that file is the following line,
```
# define NO_DATA	4	/* Valid name, no data record of requested`
```
and in `transmission_loader.h` there's this enum defined,
```
enum ParseStatus
  {
    SUCCESS,
    NO_DATA,
    BAD_TYPE
  };
```
which is a collision and you can't compile.

jlack1987:
This is to fix https://github.com/ros-controls/ros_control/issues/333. I ultimately removed the `ParseStatus` enum. It didn't appear there being 3 valid return types was adding value so I thought removing it made the most sense. Let me know what you think

### DIFF
--- a/transmission_interface/src/transmission_loader.cpp
+++ b/transmission_interface/src/transmission_loader.cpp
@@ -32,7 +32,7 @@
 namespace transmission_interface
 {
 
-TransmissionLoader::ParseStatus
+bool
 TransmissionLoader::getActuatorReduction(const TiXmlElement& parent_el,
                                          const std::string&  actuator_name,
                                          const std::string&  transmission_name,
@@ -47,13 +47,14 @@ TransmissionLoader::getActuatorReduction(const TiXmlElement& parent_el,
     {
       ROS_ERROR_STREAM_NAMED("parser", "Actuator '" << actuator_name << "' of transmission '" << transmission_name <<
                              "' does not specify the required <mechanicalReduction> element.");
+      return false;
     }
     else
     {
       ROS_DEBUG_STREAM_NAMED("parser", "Actuator '" << actuator_name << "' of transmission '" << transmission_name <<
                              "' does not specify the optional <mechanicalReduction> element.");
+      return true;
     }
-    return ParseStatus::NO_DATA;
   }
 
   // Cast to number
@@ -62,12 +63,12 @@ TransmissionLoader::getActuatorReduction(const TiXmlElement& parent_el,
   {
     ROS_ERROR_STREAM_NAMED("parser", "Actuator '" << actuator_name << "' of transmission '" << transmission_name <<
                            "' specifies the <mechanicalReduction> element, but is not a number.");
-    return ParseStatus::BAD_TYPE;
+    return false;
   }
-  return ParseStatus::SUCCESS;
+  return true;
 }
 
-TransmissionLoader::ParseStatus
+bool
 TransmissionLoader::getJointReduction(const TiXmlElement& parent_el,
                                       const std::string&  joint_name,
                                       const std::string&  transmission_name,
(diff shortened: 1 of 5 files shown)


bmagyar (approved):
The changes look good to me although I'm a little worried that this PR didn't have to touch any tests...

bmagyar:
@ipa-mdl or @matthew-reynolds for a second look?

transmission_interface/src/transmission_loader.cpp
```diff
@@ -53,7 +53,7 @@ TransmissionLoader::getActuatorReduction(const TiXmlElement& parent_el,
       ROS_DEBUG_STREAM_NAMED("parser", "Actuator '" << actuator_name << "' of transmission '" << transmission_name <<
                              "' does not specify the optional <mechanicalReduction> element.");
     }
-    return ParseStatus::NO_DATA;
+    return true;
```
matthew-reynolds: I think this should return `false`, at least in the case that `required==true`. That would make this change line up better with the `status != ParseStatus::SUCCESS`->`!status` change.

transmission_interface/src/transmission_loader.cpp
```diff
@@ -62,12 +62,12 @@ TransmissionLoader::getActuatorReduction(const TiXmlElement& parent_el,
   {
     ROS_ERROR_STREAM_NAMED("parser", "Actuator '" << actuator_name << "' of transmission '" << transmission_name <<
                            "' specifies the <mechanicalReduction> element, but is not a number.");
-    return ParseStatus::BAD_TYPE;
+    return false;
```
matthew-reynolds, line 66: I liked @ipa-mdl's suggestion in #333 of using exceptions for this case, to differentiate from the "element not found" case. Or perhaps the best solution would be:

 - An exception for "Required element not found"
 - An exception for "Element is not a number/BAD_TYPE"
 - False for "Optional element not found"
 - True for success

I'm just hesitant to lump together the error conditions.

transmission_interface/src/transmission_loader.cpp
jlack1987: yeah I think I agree w/you there. I'll make the fix

transmission_interface/src/transmission_loader.cpp
jlack1987: I'm not sure I fully agree with returning false if an optional element wasn't found. In that case the caller would need to know more information otherwise things would likely terminate even though the only thing that happened was an optional thing wasn't found.

matthew-reynolds: > the caller would need to know more information

I would argue that with the exception model I mentioned, you *would* be giving them sufficient information, since it differentiates between exceptions/an error vs false/"optional not found". Plus this is an internal API anyways

The main reason I'm pushing this is because "Optional element not found" used to return `ParseStatus::NO_DATA`, so it makes me a bit uncomfortable to change it to a silent failure.

jlack1987: yeah I think my main point of contention is that in the case of `ParseStatus::NO_DATA` it was already a "silent" failure in which you would only get a rosconsole message. Just looking at one of the callers to these functions [here](https://github.com/ros-controls/ros_control/blob/noetic-devel/transmission_interface/src/differential_transmission_loader.cpp#L234). It only actually checks if the return type is `ParseStatus::BAD_TYPE`, so if it's `ParseStatus::SUCCESS` or `ParseStatus::NO_DATA` it will simply continue on.

matthew-reynolds: Ah I see, that's fair. I hadn't noticed that every time it's called with optional, the existing callers makes a point to only check `BAD_TYPE` rather than `!SUCCESS`.

transmission_interface/src/transmission_loader.cpp
jlack1987: Ok I have update it to return false when `required == true`

transmission_interface/src/transmission_loader.cpp
jlack1987: Let me know if there is some action you'd like me to take here, otherwise I think I addressed the other comment. Thanks!

matthew-reynolds: LGTM

matthew-reynolds (approved):

jlack1987:
@bmagyar unless there are further reviewer comments incoming I think this one is ready to go

bmagyar:
It is indeed, thanks for the contribution!

PR #471, opened 2020-08-10
Use an explicit relative import path instead of implicit.

fmauch:
On python3 systems the implicit relative import will not work. The explicit
notation, however, should work on python >= 2.5

I am targeting melodic, as I myself am using ROS melodic with a python3 environment. According to [REP-0003](https://www.ros.org/reps/rep-0003.html) `noetic-devel` would probably be a more appropriate target branch, but it probably wouldn't hurt to include it into melodic, as well.


bmagyar (approved):

bmagyar:
I've cherry-picked to noetic too and pushing a new release to both noetic and melodic with this, thanks for the contribution!

PR #472, opened 2020-08-19
Remove irrelevant test case from transmission_interface

### LINKED ISSUE #460: Remove irrelevant test case from transmission_interface
We used to host an implementation of `is_permutation` which got replaced by the standard library alternative once it was out. The test for the old function however was simply refactored instead of removed. 

https://github.com/ros-controls/ros_control/blob/noetic-devel/transmission_interface/test/transmission_interface_loader_test.cpp#L63

I suggest we don't test STL functions, let's remove it!

mateus-amarante:
Resolve #460


matthew-reynolds (approved):
LGTM

bmagyar:
Thanks!

bmagyar (approved):

PR #473, opened 2020-08-19
Fix jnt-to-act command propagation check in transmission_interface_loader_test

mateus-amarante:
Looking at [transmission_interface_loader_test.cpp](https://github.com/ros-controls/ros_control/blob/noetic-devel/transmission_interface/test/transmission_interface_loader_test.cpp), I noticed the propagation of the joint commands are being tested with the joint states.


bmagyar (approved):
Looks good to me, nice catch!!

PR #475, opened 2020-09-15
re-enable controllers in source-based CI

bmagyar:


bmagyar:
@mateus-amarante success!! Thanks for working this out!

PR #478, opened 2020-09-20
Move ros_controllers to DOWNSTREAM_WORKSPACE (noetic CI)

mateus-amarante:
Since `ros_controllers` is a package that depends on `ros_control`, I think it should be tested too, but in DOWNSTREAM_WORKSPACE. If you just want to check it builds (and not run all its tests), it is possible to set `NOT_TEST_DOWNSTREAM=true`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ notifications:
 
 env:
   global:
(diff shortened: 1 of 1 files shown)


bmagyar:
ah, that's a very good idea!

Issue #507, opened 2022-12-06
Recent changes to spawner script break simulation startup for legged robots


Martin-Oehler:
Hello everyone,
this recent commit https://github.com/ros-controls/ros_control/commit/66a4ef76f4f6d5b4e3e2bcda4e4f0a3dc7bd7e4b (issue https://github.com/ros-controls/ros_control/issues/431, PR https://github.com/ros-controls/ros_control/pull/432) introduced waiting in a loop until the ros time is running. This leads to the problem, that controllers will start very late (~500 ms after simulation start, before the change: ~80 ms) and robots that rely on running controllers to be stable such as legged robots just collapse (and the simulation usually explodes). I could not find an easy work-around that works on all machines, so legged robot simulation is currently broken for us.

I think that everyone using legged robots in Gazebo with ros_control should have this problem as there is no way to start controllers faster. 

Best regards,
Martin

bmagyar:
Could you reintroduce the functionality with a feature flag that is off by default? This way everybody gets what they need

Martin-Oehler:
You mean like `--wait-for-ros-time` or `--do-not-wait-for-ros-time`?

Alternatively, `wait_for_services` could be called as normal and only when that times out, the check for sim time could happen. This is a bit more complicated but should solve both problems, I think.

Martin-Oehler:
I will prepare a PR for my proposed solution that should address both cases.

PR #508, opened 2022-12-06
Reorder waiting for services and ros time to speed up controller start

Martin-Oehler:
Hello everyone,
this PR addresses issue https://github.com/ros-controls/ros_control/issues/507. I reordered the code structure so `waiting_for_service` happens first and only if that times out, check for running sim time happens. 

Maybe @guillaumeautran could also check, if this works for them?

Best regards,
Martin


Martin-Oehler:
Could you review this please? It should fix the problems in both use cases.

bmagyar (approved):

ravikola02:
Thank you for figuring this out. Can we do this for melodic as well? @bmagyar @Martin-Oehler

PR #516, opened 2023-02-17
Reorder waiting for services and ros time to speed up controller start for Melodic branch

ravikola02:
These changes are to resolve issue #507 , which was updated for noetic in #508 but not for melodic branch.


bmagyar (approved):
I can merge this but keep in mind there won't be new releases to Melodic

PR #519, opened 2023-10-26
QTimer::setInterval expects int argument

rhaschke:
Since Ubuntu 22.04 / Qt 5.15, Qt seems to be pickier about the correct type, and loading the rqt plugin fails.
https://doc.qt.io/qt-5/qtimer.html#interval-prop

### DIFF
--- a/combined_robot_hw_tests/package.xml
+++ b/combined_robot_hw_tests/package.xml
@@ -24,6 +24,8 @@
 
   <depend>combined_robot_hw</depend>
   <depend>controller_manager</depend>
+  <depend>controller_manager_msgs</depend>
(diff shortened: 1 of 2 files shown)


bmagyar:
Seems pretty harmless for Noetic still... Am I correct?

rhaschke:
Hi Bence,
this aims for [ROS One / Obese](https://github.com/ros-o) (running on Ubuntu 22.04). But it shouldn't harm on Noetic either.

rhaschke:
I don't think the test failure is linked to my change.

bmagyar:
agreed, I'll re-run them, I think it's only flakiness

bmagyar (approved):

